Issue #21, opened 2015-01-05
Hook to reject or warn about case changes in filenames


guykisel:
In case sensitive filesystems, you could commit two files that have the same name but with different case (`myfile.txt` vs `MyFile.txt`), but then if these are checked out into a case insensitive filesystem, the results are unpredictable. A hook to warn or reject commits that result in these sorts of filenames would be nice.

asottile:
+1 this is a great idea

lmk if you want me to take a stab at this :)

guykisel:
I was going to give it a try when I have time, but it won't hurt my feelings if you do it instead!

guykisel:
This was implemented in #22

PR #22, opened 2015-01-07
Add check_case_conflict hook

guykisel:
This is for #21. I'm not 100% sure about the validity of my approach or of my test cases, so comments/criticism are welcome.

### DIFF
--- a/pre_commit_hooks/check_case_conflict.py
+++ b/pre_commit_hooks/check_case_conflict.py
@@ -1,0 +1,58 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+
+from plumbum import local
+
+from pre_commit_hooks.util import added_files
+
+
+def lower_set(iterable):
+    return set(x.lower() for x in iterable)
+
+
+def find_conflicting_filenames(filenames):
+    repo_files = set(local['git']('ls-files').splitlines())
+    relevant_files = set(filenames) | added_files()
+    repo_files -= relevant_files
+    retv = 0
+
+    # new file conflicts with existing file
+    conflicts = lower_set(repo_files) & lower_set(relevant_files)
+
+    # new file conflicts with other new file
+    lowercase_relevant_files = lower_set(relevant_files)
+    for filename in set(relevant_files):
+        if filename.lower() in lowercase_relevant_files:
+            lowercase_relevant_files.remove(filename.lower())
+        else:
+            conflicts.add(filename.lower())
+
+    if conflicts:
+        conflicting_files = [
+            x for x in repo_files | relevant_files
+            if x.lower() in conflicts
+        ]
+        for filename in sorted(conflicting_files):
+            print('Case-insensitivity conflict found: {0}'.format(filename))
+        retv = 1
+
+    return retv
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'filenames', nargs='*',
+        help='Filenames pre-commit believes are changed.'
+    )
+
+    args = parser.parse_args(argv)
+
+    return find_conflicting_filenames(args.filenames)
+
+
+if __name__ == '__main__':
+    exit(main())
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ env: # These should match the tox env list
     - TOXENV=pypy3
 install: pip install coveralls tox --use-mirrors
 script: tox
+# Special snowflake.  Our tests depend on making real commits.
+before_install:
+  - git config --global user.name "Travis CI"
+  - git config --global user.email "user@example.com"
 after_success:
     - coveralls
 sudo: false
--- a/hooks.yaml
+++ b/hooks.yaml
@@ -12,6 +12,13 @@
     language: python
     # Match all files
     files: ''
+-   id: check-case-conflict
+    name: Check for case conflicts
+    description: Check for files that would conflict in case-insensitive filesystems
+    entry: check-case-conflict
+    language: python
+    # Match all files
+    files: ''
 -   id: check-docstring-first
     name: Check docstring is first
     description: Checks a common error of defining a docstring after code.
--- a/pre_commit_hooks/check_added_large_files.py
+++ b/pre_commit_hooks/check_added_large_files.py
@@ -8,15 +8,13 @@ import math
 import os
 import sys
 
-from plumbum import local
+from pre_commit_hooks.util import added_files
 
 
 def find_large_added_files(filenames, maxkb):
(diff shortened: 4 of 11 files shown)


guykisel:
Hmm, the tests failed in Travis when I tried to commit to the temporary git repo. Not really sure what to do about that.

pre_commit_hooks/check_case_conflict.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # Compare the files pre-commit says are modified against all files in the
+    # repo
+    pairs = itertools.product(filenames, repo_filenames)
+
+    retv = 0
+    for pair in pairs:
+        if pair[0] != pair[1] and pair[0].lower() == pair[1].lower():
+            print('{0} and {1} differ only in case.'.format(pair[0], pair[1]))
+            retv = 1
```
asottile: a more efficient (and maybe more correct?) approach:

```
1. repo_files = set(`git ls-files`)
2. added_files = set((I forget, find the code in the large files thing, make it into a util that both use))
3. pre_commit_files = set((get these from argv))

relevant_files = pre_commit_files & added_files
repo_files -= relevant_files

if set(x.lower() for x in repo_files) & set(x.lower() for x in relevant_files):
    # print useful error message

```

asottile: oh you'll also want to check `len(relevant_files) == len(set(x.lower() for x in relevant_Files))`

pre_commit_hooks/check_case_conflict.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return retv
+
+
+def main(argv=None):
+    argv = argv if argv is not None else sys.argv[1:]
```
asottile: so I've found that parser.parse_args(None) will magically use sys.argv, so you can actually drop this line

asottile:
as for travis, we solved this here: https://github.com/pre-commit/pre-commit/blob/master/.travis.yml#L10

guykisel:
Cool, thanks for the suggestions!

guykisel:
Tests are passing after the tox config fix, but there are errors due to duplicate code. I'm working on those errors + the suggestions above.

guykisel:
So while I'm working on this, I've realized that the unit tests for this mostly only work on a case-sensitive filesystem. Does it seem like a bad idea to do this:
```python
def filesystem_is_case_insensitive():
    """From http://stackoverflow.com/a/7870076/2047682"""
    # By default mkstemp() creates a file with
    # a name that begins with 'tmp' (lowercase)
    tmphandle, tmppath = tempfile.mkstemp(dir=os.getcwd())
    case_insensitive = os.path.exists(tmppath.upper())
    os.remove(tmppath)
    return case_insensitive

def test_adding_something_with_conflict(temp_git_dir):
    if filesystem_is_case_insensitive():
        return
    with local.cwd(temp_git_dir):
        write_file('f.py', "print('hello world')")
        local['git']('add', 'f.py')
        write_file('F.py', "print('hello world')")
        local['git']('add', 'F.py')

        assert find_conflicting_filenames(['f.py', 'F.py']) == 1
```

guykisel:
Also, weird note about developing this on OS X (which defaults to case-insensitive) - I created a case-sensitive disk image to work in, then had to override pytest's default base temporary directory, because it was creating temporary files in the case-insensitive volume.

guykisel:
Note this is still a work in progress. My conflict detection still seems overly complicated.

guykisel:
On my macbook, the current algorithm takes ~40ms for 10k filenames, and ~1s for 1m filenames.

tox.ini
```diff
@@ -8,7 +8,7 @@ install_command = pip install --use-wheel {opts} {packages}
 deps = -rrequirements-dev.txt
 commands =
     coverage erase
-    coverage run -m pytest {posargs:tests}
+    coverage run -m pytest --basetemp=tmp {posargs:tests}
```
asottile: You can accomplish this in userspace by doing:

`TMPDIR=tmp tox -e py27` (or `TMPDIR=tmp make test`)

pre_commit_hooks/check_case_conflict.py
```diff
@@ -0,0 +1,53 @@
...
+
+def find_conflicting_filenames(filenames):
+    repo_files = set(local['git']('ls-files').splitlines())
+    relevant_files = set(filenames) | added_files()
+    repo_files -= relevant_files
+    retv = 0
+
+    # new file conflicts with existing file
+    conflicts = (set(x.lower() for x in repo_files)
+                 & set(x.lower() for x in relevant_files))
```
asottile: just a nit:

I usually prefer this written as:

conficts = (
    (...) &
    (...)
)

pre_commit_hooks/check_case_conflict.py
```diff
@@ -0,0 +1,53 @@
...
+    lowercase_relevant_files = set(x.lower() for x in relevant_files)
+    for filename in set(relevant_files):
+        if filename.lower() in lowercase_relevant_files:
+            lowercase_relevant_files.remove(filename.lower())
+        else:
+            conflicts.add(filename.lower())
+
+    if conflicts:
+        conflicting_files = [x for x in repo_files | relevant_files if
+                             x.lower() in conflicts]
```
asottile: another nit:

I'd write this as:
```
conflicting_files = [
     x for x in repo_files | relevant_files
    if x.lower() in conflicts
]
```

pre_commit_hooks/check_case_conflict.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def find_conflicting_filenames(filenames):
+    repo_files = set(local['git']('ls-files').splitlines())
+    relevant_files = set(filenames) | added_files()
+    repo_files -= relevant_files
+    retv = 0
+
+    # new file conflicts with existing file
+    conflicts = (set(x.lower() for x in repo_files)
```
asottile: seems like this code might be slightly more readable with a

```
def lower_set(iterable):
    return set(x.lower() for x in iterable)
```

asottile:
Overall seems fine.

- To fix the pylint stuff, move the `temp_git_dir` fixture into tests/conftest.py
- Remove the `--basetemp`
- Squash commits
- Might need to add this to pylintrc also:
```
[SIMILARITIES]
# maybe min-similarity-lines=N as well?
ignore-imports=yes
```
And I'll merge this :D

Thanks for the PR!

guykisel:
Fixed and squashed.

asottile:
@guykisel Thanks for the PR :D

guykisel:
:)

asottile:
Ooh one more thing, can you update the README?

guykisel:
Oh, sure. Want a new PR for that?

PR #23, opened 2015-01-12
change wording so that people know this hook needs arguments.

kentwills:
I got tripped up because this is the only hook that requires arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Add this to your `.pre-commit-config.yaml`
 
 ### Hooks available
 
-- `autopep8-wrapper` - Runs autopep8 over python source.
+- `autopep8-wrapper` - Runs autopep8 over python source. (You'll want `args: ['-i]` when using this hook, see `.pre-commit-config.yaml` for an example.)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ Add this to your `.pre-commit-config.yaml`
 
 ### Hooks available
 
-- `autopep8-wrapper` - Runs autopep8 over python source.
+- `autopep8-wrapper` - Runs autopep8 over python source. (You'll want `args: ['-i]` when using this hook, see `.pre-commit-config.yaml` for an example.)
```
asottile: should also update hooks.yaml to include this

hooks.yaml
```diff
@@ -1,6 +1,6 @@
 -   id: autopep8-wrapper
     name: autopep8 wrapper
-    description: Runs autopep8 over python source.
+    description: Runs autopep8 over python source. (You'll want args: [-i] when using this hook, see .pre-commit-config.yaml for an example.)
```
asottile: this is invalid yaml, you'll need to quote this literal (yuno use pre-commit hooks!)

PR #34, opened 2015-02-09
implemented a basic double quote string fixer

struys:


struys:
oops. totally broken, fixing now.

README.md
```diff
@@ -36,6 +36,7 @@ Add this to your `.pre-commit-config.yaml`
 - `name-tests-test` - Assert that files in tests/ end in _test.py
 - `pyflakes` - Run pyflakes on your python files
 - `requirements-txt-fixer` - Sorts entries in requirements.txt
+- `double-quote-string-fixer` - Replaces strings with double quoted with single quotes
```
asottile: copy the description from hooks.yaml because this makes no sense :)

pre_commit_hooks/string_fixer.py
```diff
@@ -0,0 +1,64 @@
...
+    string = m.group(0)
+
+    for double_quote_start in double_quote_starts:
+        if string.startswith(double_quote_start):
+            meat = string[len(double_quote_start):-1]
+            if '"' in meat or "'" in meat:
+                break
+            return (
+                double_quote_start.replace('"', "'") +
+                string[len(double_quote_start):-1] +
```
asottile: meat

pre_commit_hooks/string_fixer.py
```diff
@@ -0,0 +1,64 @@
...
+                # Docstrings are hard, fuck it
+                lines.append(line)
+            else:
+                result = re.sub(tokenize.String, handle_match, line)
+                lines.append(result)
+                return_value |= int(result != line)
+
+    with open(filename, 'w') as write_handle:
+        for line in lines:
+            write_handle.write(line)
```
asottile: Should be able to re.sub the whole file at once.

You'll also want to `re.compile` at the top of the file I think?

struys: Making the re.compile change.

If I re.sub the entire file, I can't skip over the docstring case. In handle_match it's not super easy to recognize the docstring case.

asottile: why not? shouldn't the string regex only match one-quoted strings?

struys: Token considers """ Foo """ to be a quoted string. Docstring test fails if I remove the special case.

PR #43, opened 2015-03-06
Detect OpenSSH private keys

campaul:


pre_commit_hooks/detect_private_key.py
```diff
@@ -0,0 +1,33 @@
...
+    parser.add_argument('filenames', nargs='*', help='Filenames to check')
+    args = parser.parse_args(argv)
+
+    rsa_key_files = local['grep'][
+        ('-l', 'BEGIN RSA PRIVATE KEY') + tuple(args.filenames)
+    ](retcode=None).strip().splitlines()
+
+    dsa_key_files = local['grep'][
+        ('-l', 'BEGIN DSA PRIVATE KEY') + tuple(args.filenames)
+    ](retcode=None).strip().splitlines()
```
asottile: let's use `re` for more cross-platform support

pre_commit_hooks/detect_private_key.py
```diff
@@ -0,0 +1,34 @@
...
+    args = parser.parse_args(argv)
+
+    private_key_files = []
+
+    for filename in args.filenames:
+        with io.open(filename, 'r') as f:
+            content = f.read()
+            if re.search('BEGIN RSA PRIVATE KEY', content):
+                private_key_files.append(content)
+            if re.search('BEGIN DSA PRIVATE KEY', content):
```
asottile: well in this case you don't even need re now :) `'...' in content`

PR #46, opened 2015-03-12
Add --django flag to test name checker

### LINKED ISSUE #44: name-tests-test isn't Django-friendly
https://docs.djangoproject.com/en/1.7/topics/testing/advanced/#using-different-testing-frameworks

Django's built-in test runner defaults to searching for `test*.py`. It would be nice if `name-tests-test` either allowed both the `test*.py` and `*_test.py` patterns or if there was a separate hook for the Django-style test naming pattern.

I'd be happy to submit a PR for the preferred solution/implementation.

guykisel:
This resolves #44

### DIFF
--- a/pre_commit_hooks/tests_should_end_in_test.py
+++ b/pre_commit_hooks/tests_should_end_in_test.py
@@ -1,23 +1,35 @@
 from __future__ import print_function
 
 import argparse
+import re
 import sys
 
 
 def validate_files(argv=None):
     parser = argparse.ArgumentParser()
     parser.add_argument('filenames', nargs='*')
+    parser.add_argument(
+        '--django', default=False, action='store_true',
+        help='Use Django-style test naming pattern (test*.py)'
+    )
     args = parser.parse_args(argv)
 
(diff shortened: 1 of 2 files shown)


shubhamchaudhary:
How can a user specify --django in pre-commit-config.yaml?

```yaml
-   repo: git://github.com/pre-commit/pre-commit-hooks
    sha: v0.8.0
    hooks:
    -   id: check-symlinks
    -   id: name-tests-test
    -   id: requirements-txt-fixer
```

shubhamchaudhary:
NM. Here is the solution:

```
-   repo: git://github.com/pre-commit/pre-commit-hooks
    sha: v0.8.0
    hooks:
    -   id: check-symlinks
    -   id: name-tests-test
        entry: name-tests-test --django
    -   id: requirements-txt-fixer
```

asottile:
That's one way, [but the intended way is to use `args`](http://pre-commit.com/#passing-arguments-to-hooks)

```yaml
    -   id: name-tests-test
        args: [--django]
```

PR #49, opened 2015-03-20
Only check merge conflicts on conflict commits

asottile:
@guykisel what do you think?


guykisel:
Seems very reasonable. Did anything in particular prompt this?

asottile:
@guykisel we check in markdown files a lot and devs would be grumpy if we suddenly started blocking those

guykisel:
In that case, an alternative to consider (no idea if this is better or worse) is to exclude running this hook on *.md files

guykisel:
Anyway :+1:

Issue #50, opened 2015-03-25
yapf hook


guykisel:
https://github.com/google/yapf looks like a neat tool to potentially replace autopep8. It does appear to be a bit immature, but once it stabilizes I think it could make for a good hook.

asottile:
Looks cool! Hopefully it stabilizes :)

jab:
Would you consider adding a hook for yapf now?

asottile:
Sure! Would you like to take a stab at a PR or should I?

jab:
Awesome, thanks! I'm still trying to get `pre-commit run` to work for the first time (new user; it's failing with `FileNotFoundError: [Errno 2] No such file or directory: 'virtualenv'`), so if you're up for it, that'd be great. If I get past this I might take a stab if no one beats me to it though.

asottile:
ah yeah I actually just fixed that here: https://github.com/pre-commit/pre-commit/pull/216

Lemme merge that and bump the version

asottile:
That bug should be fixed with 0.4.4 of pre-commit

jab:
Nice, thanks!

asottile:
@jab if you see additional bugs like that, be sure to report them here: https://github.com/pre-commit/pre-commit/issues :D

jab:
Will do!

Meanwhile, will keep an eye on this ticket for yapf support.

Thanks for making this project, it's awesome!

jab:
@asottile I had a chance to start a PR for this in #51. Not ready to be merged yet but figured I'd create it now anyway to get the ball rolling. In particular, looking at the autopep8 wrapper as an example, I'm not sure if we can similarly call into yapf's arg parsing code to get individual files out of the `files` arg and format them one-by-one, since yapf creates the arg parser inside its `main` function. What do you recommend?

Also just noting a test should be added once that's working.

Feel free to move discussion over to #51 if that's a better place for it.

struys:
@jab can you just call the main in a loop? Pass in individual files and print if the return code is non-0

jab:
@struys The issue is we want to parse the individual files out of argv using yapf's arg parsing code before calling its `main`, but its arg parsing code is inside its `main`. Maybe they'd be amenable to factoring it out into its own function so it could be called separately as with autopep8, but for now that's not what they do.

asottile:
Might be worth poking them to see what they think.  Maybe they'd even support our interface and add a hooks.yaml file

[...]

PR #51, opened 2015-03-30
first pass at #50 (yapf hook)

jab:


guykisel:
What does the console output of this hook look like when it's fixing files? I think the autopep8 hook might be useful for comparison purposes. It runs autopep8 individually on files, and logs filenames that it's fixing.

jab:
The console output just looks like
```
yapf wrapper..............................................................................................................................................................................Passed
```
and it would definitely be nice if it did something more like autopep8 wrapper, but see https://github.com/pre-commit/pre-commit-hooks/issues/50#issuecomment-87791248.

I actually think that the way this is currently calling yapf, the output is not even being saved.

asottile:
So now that https://github.com/pre-commit/pre-commit/pull/287 is implemented, yapf can be used directly without needing a wrapper :D

asottile:
well, we'll still need to mirror in some way, but at least we don't care so much about the exit code if `-i` is being used

jab:
original PR opener here, cool to see the progress. haven't touched any of this in months - can anyone post a quick reminder what to do to try this out?

asottile:
@jab I've set up a mirror here: https://github.com/pre-commit/mirrors-yapf

PR #54, opened 2015-04-16
Add eyaml support for check-yaml hook

rhoml:
In Puppet a common practice is to add eyaml files to hiera for security purposes.


asottile:
Thanks!

PR #55, opened 2015-05-07
Don't report markup titles as merge conflicts

dupuy:
Several markup formats, such as Markdown or Re(Structured)Text can format titles as text with '=' characters as double underlining, like this:
```
My Page Title
=============

Lorem ipsum...
```

Rather that considering any line starting with seven '=' as a conflict marker, require a space (or line-ending newline) after the equals.

This could still create a false positive for a seven character title, like "Problem", but the markup formats generally allow extra '=' characters, so by formatting the text like this:
```
Problem
========

Not...
```
these pre-commit warnings can be avoided.

### DIFF
--- a/pre_commit_hooks/check_merge_conflict.py
+++ b/pre_commit_hooks/check_merge_conflict.py
@@ -6,7 +6,8 @@ import sys
 
 CONFLICT_PATTERNS = [
     '<<<<<<< ',
-    '=======',
+    '======= ',
+    '=======\n',
     '>>>>>>> '
 ]
 WARNING_MSG = 'Merge conflict string "{0}" found in {1}:{2}'
--- a/tests/check_merge_conflict_test.py
+++ b/tests/check_merge_conflict_test.py
@@ -47,7 +47,7 @@ def f1_is_a_conflict_file(in_tmpdir):
 
 
 @pytest.mark.parametrize(
-    'failing_contents', ('<<<<<<< HEAD', '=======', '>>>>>>> master'),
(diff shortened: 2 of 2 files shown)


asottile:
:+1:
Looks good!

fixup the tests and I'll merge :)

dupuy:
Full tests are running, but early returns indicate success.

dupuy:
One final note, the names of some of the functions are a bit misleading: the check function `is_in_merge_conflict` in check_merge_conflict.py is actually detecting that the repository is currently in a merging state, not that any particular file is in a conflict state.  It should probably be renamed to `is_in_merge`, since it does have the behavior we want - Git won't allow a commit to proceed while a file is in a conflict state, so detecting that in pre-commit hook is pointless.
Similarly, the test fixture function `f1_is_a_conflict_file` does indeed make f1 a conflict file, but nothing cares about that - the point is that it leaves the repository in merging state, and adding a `@pytest.mark.usefixtures('f1_is_a_conflict_file')` before the test_does_not_care_when_not_in_a_conflict function would cause it to fail even though it does nothing with any file named `f1`.  I would suggest renaming that fixture to `repository_is_pending_merge`.

dupuy:
I wrote:
> the test fixture function f1_is_a_conflict_file does indeed make f1 a conflict file, but nothing cares about that - ... I would suggest renaming that fixture to repository_is_pending_merge.

Rather than renaming, in [PR#56](https://github.com/pre-commit/pre-commit-hooks/pull/56) I kept it but added a separate repository_is_pending_merge fixture that uses `git pull --no-commit` to leave the repository in non-conflicted merge-pending state, and added some meta-assertions about git behavior that the detection code is relying on, in case that should ever change or go away.

PR #58, opened 2015-05-10
Implement Markdown trailing space line break preservation

dupuy:
Markdown uses two or more trailing spaces on a line to indicate a forced
line break `<br/>` - these will be preserved for files with a markdown
extension (default = `.md` or `.markdown`).
Add `--markdown-linebreak-ext=X,Y` to set extensions (`*` matches any),
and `--no-markdown-linebreak-ext` to disable this feature.

Update README.md to include information on these arguments as well as
arguments added to other hooks

### DIFF
--- a/pre_commit_hooks/trailing_whitespace_fixer.py
+++ b/pre_commit_hooks/trailing_whitespace_fixer.py
@@ -2,18 +2,44 @@ from __future__ import print_function
 
 import argparse
 import fileinput
+import os
 import sys
 
 from pre_commit_hooks.util import cmd_output
 
 
-def _fix_file(filename):
+def _fix_file(filename, markdown=False):
     for line in fileinput.input([filename], inplace=True):
+        # preserve trailing two-space for non-blank lines in markdown files
+        if markdown and (not line.isspace()) and (line.endswith("  \n")):
+            line = line.rstrip(' \n')
+            # only preserve if there are no trailing tabs or unusual whitespace
+            if not line[-1].isspace():
+                print(line + "  ")
+                continue
+
         print(line.rstrip())
 
 
 def fix_trailing_whitespace(argv=None):
     parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--no-markdown-linebreak-ext',
+        action='store_const',
+        const=[],
+        default=argparse.SUPPRESS,
+        dest='markdown_linebreak_ext',
+        help='Do not preserve linebreak spaces in Markdown'
+    )
+    parser.add_argument(
+        '--markdown-linebreak-ext',
+        action='append',
+        const='',
+        default=['md,markdown'],
+        metavar='*|EXT[,EXT,...]',
+        nargs='?',
+        help='Markdown extensions (or *) for linebreak spaces'
+    )
     parser.add_argument('filenames', nargs='*', help='Filenames to fix')
     args = parser.parse_args(argv)
 
@@ -21,10 +47,28 @@ def fix_trailing_whitespace(argv=None):
         'grep', '-l', '[[:space:]]$', *args.filenames, retcode=None
     ).strip().splitlines()
 
+    md_args = args.markdown_linebreak_ext
+    if '' in md_args:
+        parser.error('--markdown-linebreak-ext requires a non-empty argument')
+    all_markdown = '*' in md_args
+    # normalize all extensions; split at ',', lowercase, and force 1 leading '.'
+    md_exts = ['.' + x.lower().lstrip('.')
+               for x in ','.join(md_args).split(',')]
+
+    # reject probable "eaten" filename as extension (skip leading '.' with [1:])
+    for ext in md_exts:
+        if any(c in ext[1:] for c in r'./\:'):
+            parser.error(
+                "bad --markdown-linebreak-ext extension '{0}' (has . / \\ :)\n"
+                "  (probably filename; use '--markdown-linebreak-ext=EXT')"
+                .format(ext)
+            )
+
     if bad_whitespace_files:
         for bad_whitespace_file in bad_whitespace_files:
             print('Fixing {0}'.format(bad_whitespace_file))
-            _fix_file(bad_whitespace_file)
+            _, extension = os.path.splitext(bad_whitespace_file.lower())
+            _fix_file(bad_whitespace_file, all_markdown or extension in md_exts)
         return 1
     else:
         return 0
--- a/tests/trailing_whitespace_fixer_test.py
+++ b/tests/trailing_whitespace_fixer_test.py
@@ -1,6 +1,10 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import sys
+
+import pytest
+
 from pre_commit_hooks.trailing_whitespace_fixer import fix_trailing_whitespace
 from testing.util import cwd
 
@@ -12,7 +16,7 @@ def test_fixes_trailing_whitespace(tmpdir):
                 ('bar.py', 'bar\t\nbaz\t\n'),
         ):
             with open(filename, 'w') as file_obj:
-                file_obj.write(contents)  # pragma: no cover (26 coverage bug)
+                file_obj.write(contents)  # pragma: no branch (26 coverage bug)
 
         ret = fix_trailing_whitespace(['foo.py', 'bar.py'])
         assert ret == 1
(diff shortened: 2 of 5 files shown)


pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -3,17 +3,33 @@
...
 
 def fix_trailing_whitespace(argv=None):
     parser = argparse.ArgumentParser()
+    parser.add_argument('--no-markdown-linebreak-ext', action='store_const',
+                        const=[], default=argparse.SUPPRESS,
+                        dest='markdown_linebreak_ext',
+                        help='Do not preserve linebreak spaces in Markdown')
+    parser.add_argument('--markdown-linebreak-ext', action='append', const='*',
+                        default=argparse.SUPPRESS, metavar='EXTS', nargs='?',
+                        help='Markdown extensions (or *) for linebreak spaces')
```
asottile: I think we can actually simplify this to one argument:

```
parser.add_argument(
    '--markdown-trailing-whitespace-extension',
    action='append',
    help='...',
)
```

The default (empty list) would mean nothing is treated as markdown, you would add more extensions by specifying the argument more than once, and the string `'*'` would be the special case where all are treated as markdown (though why support this? I'm not sure anyone will use the `'*'` option?)

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -21,10 +37,19 @@ def fix_trailing_whitespace(argv=None):
         'grep', '-l', '[[:space:]]$', *args.filenames, retcode=None
     ).strip().splitlines()
 
+    # combine all extension arguments, splitting at ',' and normalizing them
+    # (lowercase and remove unnecessary leading '.' which may be present)
+    md_args = vars(args).get('markdown_linebreak_ext', ['md', 'markdown'])
```
asottile: `vars(...).get(` => `getattr(` usually

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -21,10 +37,19 @@ def fix_trailing_whitespace(argv=None):
...
+    md_exts = [x.lower().lstrip('.') for x in ','.join(md_args).split(',')]
+    all_markdown = '*' in md_exts
+
     if bad_whitespace_files:
         for bad_whitespace_file in bad_whitespace_files:
+            # get extension ([1]); remove leading '.' ([1:]); and make lowercase
+            extension = os.path.splitext(bad_whitespace_file)[1][1:].lower()
```
asottile: might want a testcase with a file that has no extension?

tests/trailing_whitespace_fixer_test.py
```diff
@@ -24,5 +24,87 @@ def test_fixes_trailing_whitespace(tmpdir):
...
 
 
+def test_fixes_trailing_markdown_whitespace(tmpdir):
+    with cwd(tmpdir.strpath):
+        for filename, contents in (
+                ('foo.md', 'foo  \nbar \n  '),
+                ('bar.markdown', 'bar   \nbaz\t\n\t\n'),
+        ):
+            with open(filename, 'w') as file_obj:
+                file_obj.write(contents)  # pragma: no cover (26 coverage bug)
```
asottile: ooh I've found that this can be replaced with `# pragma: no branch` usually

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -3,17 +3,33 @@
...
+
         print(line.rstrip())
 
 
 def fix_trailing_whitespace(argv=None):
     parser = argparse.ArgumentParser()
+    parser.add_argument('--no-markdown-linebreak-ext', action='store_const',
+                        const=[], default=argparse.SUPPRESS,
+                        dest='markdown_linebreak_ext',
+                        help='Do not preserve linebreak spaces in Markdown')
```
asottile: nitpick:

At least in our repos we prefer this indentation style:

```
foo(
    bar,
    baz,
    womp,
)
```

Since it:
- avoids arbitrary indentation
- avoids merge conflicts when adding arguments to function calls

Not a big deal though, I'll fixup when merging

pre_commit_hooks/trailing_whitespace_fixer.py
asottile: another idea, since in both places you're normalizing to not-dot, you could instead normalize to dotted-always in just one place (might avoid the no-extension case.  If that's the case I'd suggest changing this code to:

```
_, extension = ...splitext(...)
```

which is slightly less error prone and slightly more self documenting

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -3,17 +3,33 @@
...
-def _fix_file(filename):
+def _fix_file(filename, markdown=False):
     for line in fileinput.input([filename], inplace=True):
+        # preserve trailing two-space for non-blank lines in markdown files
+        if markdown and (not line.isspace()) and (line.endswith("  \n")):
+            line = line.rstrip(' \n')
+            # only preserve if there are no trailing tabs or unusual whitespace
+            if not line[-1].isspace():
+                print(line + "  ")
+                continue
```
asottile:
```
loop:
    if ...:
        ...1
        continue
    ...2
```

is more clearly written:

```
loop:
    if ...:
        ...1
    else:
        ...2
```

dupuy:
Thanks for the detailed code review.  It's not something I get in my day job. :-(

To simplify tracking the revisions, I have added individual commits to address each issue separately - I will squash them once we have consensus on the final version. (It's hard to follow a line note conversation when the underlying code is rebased).

Rather than reply to all the line notes, I'll combine all the ones I agree with here:

> At least in our repos we prefer this indentation style:
> ```
foo(
    bar,
    baz,
    womp,
)

[Done](https://github.com/dupuy/pre-commit-hooks/commit/b7bb0a7e27ed7a6a728a7c8179802cf55650b785).

> `vars(...).get(` => `getattr(` usually

Thanks, that is much better - I had started with the dict vars approach and didn’t re-think that. [Done](https://github.com/dupuy/pre-commit-hooks/commit/7d431fa42078f12b3e53c77b0ffc7b657be12dd8).

> might want a testcase with a file that has no extension?

Excellent suggestion. [Done](https://github.com/dupuy/pre-commit-hooks/commit/9f21b78c7baee758dfb46deb2493f5c3f2ee92c5).
I also added testcases with dotfiles (i.e. `/x/y/.foobar` should *not* match 'foobar' extension, and ended up [refactoring extensively](https://github.com/dupuy/pre-commit-hooks/commit/2504b21a3c70536abd81efcd0ad91dd8c07f6030) to use pytest.mark.parametrize for all the input filename, contents, and expected output.

> > `file_obj.write(contents)  # pragma: no cover (26 coverage bug)`

> I've found that this can be replaced with # pragma: no branch usually

I was just copying this from the existing test but you are right. [Done](https://github.com/dupuy/pre-commit-hooks/commit/f5dbfb06a945bf4238cdf9940df3c79e2356181d).

pre_commit_hooks/trailing_whitespace_fixer.py
dupuy: I really think the best default is for the common extensions (`.md` and `.markdown`) to get the special treatment.  While this is a bit more of an edge case that replacing tabs with spaces in a Makefile, I still think it is better to come down on the side of safety (i.e. no change to semantics) even if it means a few non-markdown `.md` files might have extraneous whitespace left in them, since at the end of the day that is a purely aesthetic consideration.  I don't expect many (any?) people to actually set these options, so the default settings are for me the most important.

Having a non-empty default requires a `store` action to clear the set.  I used an `append` action to support somebody who adds extensions one at a time, but my original code was not consistent as the first option overwrote the defaults rather than appending to them; that is too much magic. It is simpler, clearer, and safer to always append, so I made [that change instead](https://github.com/dupuy/pre-commit-hooks/commit/cc29df5d4544955713f58f5b71fc895ec7d7ca73).

The other reason for a separate `--no-markdown-linebreak-ext` option with a `store` action (rather than using a single option with a `store` action to set (or if empty, clear) the extension list) is that argparse will silently "eat" a following non-option argument (interpreting a staged filename as an extension) if you omit the `=` (e.g. `--markdown-linebreak-ext` instead of `--markdown-linebreak-ext=`).  I actually ran into that during testing, thus the `--` argument and comment in `test_markdown_linebreak_ext_opt_all`.  It's better to have the code error out if the argument provided is missing (`--` or other following option), empty (`=` or explicit empty string argument), or has a non-initial `.` (probably an innocent filename) so I [made that change too](https://github.com/dupuy/pre-commit-hooks/commit/00eb809a2518b2a2e1c99f951a1d86b97c9c2239).

Probably the most common case for the special `*` extension would be to enable linebreak preservation for files without extensions (you could use `--markdown-linebreak-ext=,` but that's pretty obscure, and I doubt anybody could figure out what that does/is supposed to do without looking at the code).

Finally, in your example you wrote `--markdown-trailing-whitespace-extension`.
I feel like the `--markdown-linebreak-ext` is already too long, but if you think the name should be changed, I'm happy to do so.

pre_commit_hooks/trailing_whitespace_fixer.py
dupuy: One of the reasons I added the normalization was to be forgiving in the case where somebody uses `--markdown-linebreak-ext=.foo` (rather than `=foo`).  Adding a `.` only if there is not one already is trickier than removing a `.` if present (lstrip), and I certainly don't want to require `.` in the option argument as `--markdown-linebreak-ext=.foo,.bar` is pretty ugly.

It's true that the current code is not self-documenting, but that is why I wrote a comment.

pre_commit_hooks/trailing_whitespace_fixer.py
dupuy: I used the `continue` to avoid repeating code. Since there are nested `if` statements (with a required rstrip and assignment between the two conditions, making it impractical to combine the two if statements using `and`), the alternatives are actually
```
loop:
    if ...:
        ...0
        if ...:
            ...1
            continue
    ...2
```
or
```
loop:
    if ...:
        ...0
        if ...:
            ...1
        else:
            ...2
    else:
        ...2
```
In this case, the repeated code is only `print(line)` so not such a big deal, let me know if you prefer that to the current use of `continue` and I'll change it.

asottile: ah I see, I missed the nested ifs, disregard me :)

pre_commit_hooks/trailing_whitespace_fixer.py
asottile: This should be enough to simplify it a bit:

```diff
-    md_exts = [x.lower().lstrip('.') for x in ','.join(md_args).split(',')]
+    md_exts = ['.' + x.lower().lstrip('.') for x in ','.join(md_args).split(',')]
```

The check for contained dots then becomes a check for `.count(2)`

pre_commit_hooks/trailing_whitespace_fixer.py
asottile: Ah that's fair.  This is acceptable as is then (can disregard my suggestion).

asottile:
This is mergeable in the current state, let me know if the dot refactor is too complicated and when you've prepped for merge :)

dupuy:
I did the dot refactoring (one `[1:]` was still needed for the "missing argument eats filename" check - see the [pre-squashed commit](https://github.com/dupuy/pre-commit-hooks/commit/f0cd9149faffc9fa5728a21f2121bb47f3f5c9df)) and I also improved that check to reject Windows filename specials `\` and `:` as well; and finally squashed everything into one commit on the PR branch.

asottile:
Thanks again!

PR #60, opened 2015-06-01
AWS credential checking

arahayrabedian:
add aws credential checking ONLY FOR YOUR OWN **SECRET** credentials if they're set in a configurable credentials file (AWS CLI tools' format), access keys are ignored.

I haven't had time to figure out how to configure the arguments from a .pre-commit-config yaml file, some help would be nice so I can update the readme further.

Let me know if anything's missing in order for this to get accepted.

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -2,26 +2,25 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import argparse
-import ConfigParser
 import os
+from six.moves import configparser
 
 
-def get_your_keys(credentials_file, ignore_access_key=False):
-    """ reads the keys in your credentials file in order to be able to look
+def get_your_keys(credentials_file):
+    """ reads the secret keys in your credentials file in order to be able to look
     for them in the submitted code.
     """
     aws_credentials_file_path = os.path.expanduser(credentials_file)
     if not os.path.exists(aws_credentials_file_path):
         exit(2)
 
-    parser = ConfigParser.ConfigParser()
+    parser = configparser.ConfigParser()
     parser.read(aws_credentials_file_path)
 
     keys = set()
     for section in parser.sections():
-        if not ignore_access_key:
-            keys.add(parser.get(section, 'aws_access_key_id'))
         keys.add(parser.get(section, 'aws_secret_access_key'))
+    print(str(keys))
     return keys
 
 
@@ -29,9 +28,9 @@ def check_file_for_aws_keys(filename, keys):
     with open(filename, 'r') as content:
         # naively match the entire file, chances be so slim
         # of random characters matching your flipping key.
-        for line in content:
-            if any(key in line for key in keys):
-                return 1
+        text_body = content.read()
+        if any(key in text_body for key in keys):
+            return 1
     return 0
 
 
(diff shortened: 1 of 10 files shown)


asottile:
Sounds like a great idea!  Write some tests and fix the comments and I think it'll be good for merge

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import ConfigParser
```
asottile: py2/py3 issue.  Feel free to add `six` as a dependency and `from six.moves import configparser`

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+        keys.add(parser.get(section, 'aws_secret_access_key'))
+    return keys
+
+
+def check_file_for_aws_keys(filename, keys):
+    with open(filename, 'r') as content:
+        # naively match the entire file, chances be so slim
+        # of random characters matching your flipping key.
+        for line in content:
+            if any(key in line for key in keys):
```
asottile:
```python
if key in content.read():
    return 1
```
?

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+        default='~/.aws/credentials',
+        help="location of aws credentials file from which to get the keys "
+              "we're looking for",
+    )
+    parser.add_argument(
+        "--ignore-access-key", 
+        action='store_true',
+        help="if you would like to ignore access keys, as there is "
+        "occasionally legitimate use for these.",
+    )
```
asottile: I think instead they'd either: 1. not use the hook or 2. use `SKIP=detect-aws-credentials git commit ...` (http://pre-commit.com/#advanced)

arahayrabedian: yeah I had seen the option to skip tests, I was thinking about a case where one may want the access key in the code and not the secret but then I realised you'd likely use env vars either way. will drop this option.

pre_commit_hooks/detect_aws_credentials.py
arahayrabedian: fair point, don't practice file reading as often as I should :P

pre_commit_hooks/detect_aws_credentials.py
arahayrabedian: this one i'll have to play around with, never used six myself but i've seen it all around me. will experiment.

arahayrabedian:
will get around to finishing this off over the weekend i guess.

arahayrabedian:
@asottile i had some issues with pylint and six, couldn't figure it, dunno if it'll happen on the CI server but I was hoping you'd have some pointers on that... all I could find was [this bug](https://bitbucket.org/logilab/pylint/issue/223/ignored-modules-should-turn-no-name-in)

asottile:
@arahayrabedian feel free to `# pylint:disabled= ...` for that file

arahayrabedian:
@asottile anything I need to do to get this in?

asottile:
Oops, sorry I didn't notice your update, I'll review now

arahayrabedian:
no worries, and dankew!

asottile:
Please PR http://github.com/pre-commit/pre-commit.github.io to add your name! <3

PR #65, opened 2015-06-10
Add new hook pretty-format-json

LeoCavaille:
This new hook allows to standardize one's JSON files (sorted key/X spaces indent).
By default it just fails if any file is not complying with the standard,
but you can also pass the arg `--autofix` and the hook will
pretty-format the file itself.

Good in use combined with the `check-json` hook.


pre_commit_hooks/pretty_format_json.py
```diff
@@ -0,0 +1,54 @@
...
+    for json_file in args.filenames:
+        try:
+            f = open(json_file, 'r')
+            contents = f.read()
+            f.close()
+
+            pretty_contents = simplejson.dumps(
+                simplejson.loads(contents),
+                sort_keys=True,
+                indent=4
```
bwbaugh: Would it be possible to make this configurable?

Our project is using 2 spaces since that's the output of most JSON tools, such as `jq`.

LeoCavaille: Sounds good. Also working on adding proper testing.

LeoCavaille:
@bwbaugh implemented. Tests should also be OK now.

asottile:
Looks good, I think there's issues with the newlines on windows maybe? fixnship

LeoCavaille:
@asottile changed to a common `\n` on all platforms, thanks.

PR #76, opened 2015-08-02
detect_private_key: print *filenames*, not the key

chriskuehl:
Currently it prints *the entire content* of the problem file (which might be really long, e.g. a Python test file with embedded dummy key) but doesn't actually print the file name.


asottile:
<3

PR #77, opened 2015-08-04
Add a hook to verify python ast.

asottile:
We want to use this as follows:

```yaml
-   id: check-ast
    language_version: python3.4
```
to "encourage" people to write python3 compatible syntax in our currently-python2.7 codebase

### DIFF
--- a/pre_commit_hooks/check_ast.py
+++ b/pre_commit_hooks/check_ast.py
@@ -1,0 +1,36 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import ast
+import os.path
+import sys
+import traceback
+
+
+def check_ast(argv=None):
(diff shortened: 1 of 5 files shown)


pre_commit_hooks/check_ast.py
```diff
@@ -0,0 +1,36 @@
...
+    parser.add_argument('filenames', nargs='*')
+    args = parser.parse_args(argv)
+
+    _, interpreter = os.path.split(sys.executable)
+
+    retval = 0
+    for filename in args.filenames:
+
+        try:
+            ast.parse(open(filename, 'rb').read(), filename=filename)
```
chriskuehl: any point in using a context manager on the `open` here?

asottile: not really, I exhaust the entire file immediately (and cpython'll just close the file right there)

bukzor:
:shipit:

PR #85, opened 2015-10-28
Show names of files containing aws credentials

deanwilson:
This change causes the name of the files containing the aws credentials to be printed when the pre-commit hooks are run. It's heavily based on the Detect Private Key code.


asottile:
Thanks!

PR #87, opened 2015-11-13
Add encoding pragma hook.  Resolves pre-commit/pre-commit#15

asottile:


pre_commit_hooks/fix_encoding_pragma.py
```diff
@@ -0,0 +1,75 @@
...
+        f.write(expected_pragma + second_line + old)
+    else:
+        f.write(expected_pragma + first_line + second_line + old)
+
+    return 1
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser('Fixes the encoding pragma of python files')
+    parser.add_argument('filenames', nargs='*', help='Filenames to fix')
```
chriskuehl-test: is `nargs='+'` better to get an error with no args?

chriskuehl-test: oops wrong account

Issue #88, opened 2015-11-19
Django test file names don't work with directory path


phoxelua:
For Django tests file names, **tests_should_end_in_test.py** will mistakenly think file names such as 'app/blah/tests/bad_django_test_fname.py' as valid django test names since the regex ` test_name_pattern = 'test.*.py'` matches anything between between 'test' and '.py'- even '/' and other unintended characters.

asottile:
Closed via #89

PR #89, opened 2015-11-19
Fixed regex matching for django test file names

phoxelua:
Addresses issue at https://github.com/pre-commit/pre-commit-hooks/issues/88, where the regex for django mistook the directory as part of the filename and, as a result failed to flag bad filenames.  

Implementation itself, strips off the directory path, leaving just the basename so that we don't have to run endswith. Also, re.match() wouldn't have worked with directories such as 'foo/tests/bar' anyways since it checks at the **beginning** of the string.


asottile:
Thanks for the fix!  Regexes are tricky

PR #90, opened 2015-12-01
Reformatted --diff-filter git argument

cchristous:
This makes the command compatible with older versions of git.


asottile:
Thanks for the PR!

cchristous:
Thanks for the amazingly quick merge!

Issue #82, opened 2015-10-18
check-large-files-added doesn't work with LFS


JannKleen:
when adding large files to git LFS, the check-large-files-added hook still complains. I think it would be best to add a flag to ignore LFS files, because this plugin is specifically used to solve the problem of adding large files to a git repo

asottile:
Is there an easy way to detect this?  I don't know much about LFS :)

JannKleen:
i think ```git lfs status --porcelain``` should give back all the information needed

asottile:
via #96

PR #96, opened 2015-12-25
Teach check-large-files-added about git-lfs.  Reslves #82.

asottile:
Works pretty well I think.

I ran into https://github.com/github/git-lfs/issues/913 and https://github.com/github/git-lfs/issues/914 while implementing, but it should mostly work.

@JannKleen it took me a while but I finally got around to this :)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,7 @@ omit =
     /usr/*
     */tmp*
     setup.py
+    get-git-lfs.py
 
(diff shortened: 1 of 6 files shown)


struys:
lg2m

bukzor:
looks legit :shipit:

JannKleen:
Works great! Thanks a lot :smile:

PR #105, opened 2016-03-12
Addding no-sort-keys to pretty_format_json

slariviere:
This allows to disable the sort on the keys.

Still works the same as before if the new argument is not specified.

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -6,10 +6,10 @@ import sys
 import simplejson
 
 
-def _get_pretty_format(contents, indent):
+def _get_pretty_format(contents, indent, no_sort_keys):
     return simplejson.dumps(
         simplejson.loads(contents),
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/pretty_format_json.py
```diff
@@ -6,10 +6,10 @@
 import simplejson
 
 
-def _get_pretty_format(contents, indent):
+def _get_pretty_format(contents, indent, no_sort_keys):
     return simplejson.dumps(
         simplejson.loads(contents),
-        sort_keys=True,
+        sort_keys=no_sort_keys,
```
asottile: the problem with this is you'll now get non-deterministic rearranging of keys:

```python
>>> json.dumps(json.loads('{"a": 1, "b": 2, "c": 3}'))
'{"a": 1, "c": 3, "b": 2}'
```

I think you'll need to load as an ordered type (`OrderedDict`?) and dump again to retain roundtrip ordering

PR #110, opened 2016-04-21
Enable support for prospector

kouk:
https://prospector.readthedocs.org/en/master/


guykisel:
With this, we could probably delete https://github.com/pre-commit/pre-commit.github.io/blob/b8a9ec65d4a5d192391c1d555634c1ad2237134e/all-repos.yaml#L17

asottile:
Without prospector being installed I don't think this will work?  I'd rather use @guykisel's mirror either way:

```yaml
-   repo: https://github.com/guykisel/prospector-mirror
    sha: 00fbd801
    hooks:
    -   id: prospector
```

kouk:
Hi guys, didn't know about that repo. Makes sense to keep using that I suppose.

asottile:
Sounds great!  Here's a list of some other great hooks as well: http://pre-commit.com/hooks.html

PR #120, opened 2016-06-10
Refactor legacy `indent: int` -> `indent: str`

sanmai-NL:
The type of `indent` parameter is preferably `str`. See http://simplejson.readthedocs.io/en/latest/index.html?highlight=dump#simplejson.dump . This change allows to specify TABs as indentation delimiter to `pretty_format_json`.

This is a breaking change given: invoking the plugin as script (as opposed to from the pre-commit config) ∧ supplying an integer value for `--indent`. `argparse` cannot discern integers from strings as provided on the command line and `simplejson.dump` as well as `json.dump` write non-well-formed JSON data when the `indent` argument is `'2'`. (Findings based on CPython 3.5.1 standard library).


sanmai-NL:
The Travis CI infra is not in working order.

tests/pretty_format_json_test.py
```diff
@@ -26,6 +26,17 @@ def test_unsorted_pretty_format_json(filename, expected_retval):
     assert ret == expected_retval
 
 
+@pytest.mark.parametrize(('filename', 'expected_retval'), (
+        ('not_pretty_formatted_json.json', 1),
+        ('unsorted_pretty_formatted_json.json', 1),
+        ('pretty_formatted_json.json', 1),
+        ('tab_pretty_formatted_json.json', 0),
+))
+def test_unsorted_pretty_format_json(filename, expected_retval):
```
asottile: this masks the name above it

asottile:
Can we do this in a way that is backwards compatible?  You can write a custom type:

```python
>>> def parse_indent(s):
...     try:
...         return int(s) * ' '
...     except ValueError:
...         return s
... 
>>> import argparse
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('foo', type=parse_indent)
_StoreAction(option_strings=[], dest='foo', nargs=None, const=None, default=None, type=<function parse_indent at 0x1244230>, choices=None, help=None, metavar=None)
>>> parser.parse_args('2')
Namespace(foo='  ')
>>> parser.parse_args('\t')
Namespace(foo='\t')
```

tests/pretty_format_json_test.py
asottile: (which is why travis-ci was failing, because this name clobbers the other one so the code in the body of the other function was not running, causing coverage to be not-100)

sanmai-NL: Thanks, fixed!

sanmai-NL:
I considered doing that, but figured if it's not needed or asked for (yet), then why do it. Deemed it not needed since this `pre-commit-hooks` repo is 
1. always being depended on against a fixed revision in the pre-commit configuration file, 
2. the new semantics of the indent delimiter only cause a problem under very specific circumstances, and
3. when it goes wrong that would be revealed by early failure later on (clearly non-well-formed JSON).

I've incorporated the validator function now, and added a deprecation warning so it's made clearer what `indent` usage is preferred in future pre-commit configurations.

asottile:
I disagree with the deprecation -- we're really just mimicking the `indent` parameter of dumps.

Minor thing also: https://github.com/pre-commit/pre-commit/blob/d845ec6e364f507b38098533edfbc0554661a01f/CONTRIBUTING.md#multi-line-method-invocation

sanmai-NL:
Okay, since the API docs for the `dump` function state that the `int` logic has remained for backward compatibility, I took it that we'd better deprecate that in favor of the new `str` logic. I have now removed that.

What specification of the `indent` parameter should we adhere to?

[`simplejson`](http://simplejson.readthedocs.io/en/latest/index.html?highlight=indent)
> For backwards compatibility with versions of simplejson earlier than 2.1.0, an integer is also accepted and is converted to a string with that many spaces.

No word about negative `int`s for `indent`.

[CPython 3.5.1, Standard Library `json`](https://docs.python.org/3/library/json.html#json.dump)
> If indent is a non-negative integer or string, then JSON array elements and object members will be pretty-printed with that indent level. An indent level of 0, negative, or "" will only insert newlines. None (the default) selects the most compact representation. Using a positive integer indent indents that many spaces per level. If indent is a string (such as "\t"), that string is used to indent each level.

I think the standard library spec does not make sense here, as I find it unlikely and unnecessary that API users would deliberately specify a negative `int` as `indent`. So I kept the exception upon trying to do that, but added the capability to use `0` indentation.

I also cleaned up the style per your comments. `flake8 pre_commit_hooks/pretty_format_json.py` didn't complain about that multiline method invocation convention though. Is there an automatic way to catch such issues?

asottile:
Unfortunately I haven't found a linter that helps enforce that yet :(  Still looking though!

Looks good now, I'll merge once travis-ci comes back :)

sanmai-NL:
Great!
Should I rebase first?

asottile:
up to you, I don't mind terribly much.  Looks like travis still isn't happy though :(

sanmai-NL:
Aargh, misformatted one line. It's good now!

asottile:
Awesome!  Thanks for the PR :D

pre_commit_hooks/pretty_format_json.py
```diff
@@ -24,6 +24,25 @@ def _autofix(filename, new_contents):
...
+def parse_indent(s):
+    # type: (str) -> str
+    try:
+        int_indentation_spec = int(s)
+        if int_indentation_spec >= 0:
+            return int_indentation_spec * ' '
+        else:
+            raise ValueError(
+                'Negative integer supplied to construct JSON indentation delimiter. ',
+            )
```
chriskuehl: is this exception ever actually propagating out? seems it's just going to get clobbered below (and you'll only see the message on python3, which also shows the original exception)

sanmai-NL: Thanks, a `ValueError` was raised regardless, but the error message was misleading in case of invalid input. I have updated my branch. Can you revert the merge and remarge with the tip of my branch?

asottile:
Let's just make another PR, I don't think it's worth a revert for an exception message :)

sanmai-NL:
Done, see #122.

PR #121, opened 2016-06-12
Increase robustness of tests involving `git`

sanmai-NL:
1. Disable automatic `git commit` GPG-signing, since that requires interaction. This issue was encountered in practice by me, causing spurious test failures.
2. In case path operands could turn out to start with dashes, escape the operand list with `--`.


asottile:
Thanks!

PR #123, opened 2016-06-20
Update debug_statement_hook.py

Oscar-Garcia:
Adds celery rdb as debugger: http://docs.celeryproject.org/en/latest/tutorials/debugging.html


asottile:
Thanks!

PR #127, opened 2016-07-02
handle editable requirements properly

vinayinvicible:
Requirement fixer does not handles the requirements of the format `-e git+ssh://git_url@tag#egg=req` properly.
This results in these kind of requirements always ordered to the file top.
This is not in sync with the output given by `pip freeze`.

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -10,6 +10,19 @@ class Requirement(object):
         self.value = None
         self.comments = []
 
+    @property
+    def name(self):
+        if self.value is None:
+            return
+
+        if self.value == b'\n':
+            return
+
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -10,6 +10,19 @@ def __init__(self):
...
         self.comments = []
 
+    @property
+    def name(self):
+        if self.value is None:
+            return
+
+        if self.value == b'\n':
+            return
+
```
asottile: I think (and I believe the travis-ci output confirms) that the code from here up is dead (likely handled by the explicit checks in `__lt__` below)

asottile:
+1, fix the coverage issue and :shipit:

asottile:
Thanks!

PR #129, opened 2016-07-12
Update setup.py

vikalpj:
Use latest flake8 to support `# noqa: <error>`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     packages=find_packages('.', exclude=('tests*', 'testing*')),
     install_requires=[
         # quickfix to prevent pep8 conflicts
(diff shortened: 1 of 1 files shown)


asottile:
We're neither going to pin nor pin to a beta version.  At best this will need to use `>=`

vikalpj:
@asottile Sounds good I will update the PR when stable version is released :)

asottile:
As a workaround you could put `additional_dependencies: ['flake8==...']` in your .pre-commit-config.yaml

vikalpj:
aha... thanks for the tip... Let me try that 👍

vikalpj:
@asottile closing for now will reopen when stable version is released :)

PR #132, opened 2016-08-12
Use py.path in one more place

asottile:


tests/check_merge_conflict_test.py
```diff
@@ -5,11 +5,11 @@
 import shutil
 
 import pytest
+from py.path import local
```
chriskuehl: i really hate py.path's naming :(

asottile: me too, also I don't think this is shippable because it breaks windows.  Stupid default newline handling. marley lod py.path

chriskuehl: `py.path: ಠ_ಠ`

PR #131, opened 2016-08-05
Fix flake8-respect-noqa AttributeError

ushuz:
`flake8-respect-noqa` 会导致以下报错，详见 https://travis-ci.com/xiachufang/xiachufang/builds/29409317 ，暂时移除 `flake8-respect-noqa` 否则 CI 会受影响

    AttributeError: Values instance has no attribute 'benchmark_keys'

另外从 upstream 更新了一下，目前跟 upstream 的 master 是同步的，同步完之后发现好像不需要把 `open` 都改成 `io.open()`，不会报跟 encoding 相关的问题了。

PR #133, opened 2016-08-12
Allow encoding pragma to be customizable

asottile:
See also #131.  CC @ushuz


ushuz:
Nice one, thank you!

chriskuehl:
:shipit:

PR #135, opened 2016-08-18
trailing-whitespace hook: restoring original file in case of failure - fixes #134

### LINKED ISSUE #134: Non-utf8 file will cause trailing-whitespace to truncate the file contents
Hi,

(I started investigating this in https://github.com/pre-commit/pre-commit/issues/397)

Bug reproductions steps, after setting up a git repository with the "trailing-whitespace" pre-commit hook installed (with the latest sha: 775a7906cd6004bec2401a85ed16d5b3540d2f94):
```
echo -en '\x3c\x61\x3e\xe9\x20\x0a\x3c\x2f\x61\x3e' > minimal_crasher.xml
git add minimal_crasher.xml
pre-commit run trailing-whitespace --files minimal_crasher.xml
```

Results in:
````
Trim Trailing Whitespace...............................................................................................................................................................Failed
hookid: trailing-whitespace

Files were modified by this hook. Additional output:

Fixing minimal_crasher.xml
  File "/home/lucas_cimon/.pre-commit/repok3wjweh4/py_env-default/bin/trailing-whitespace-fixer", line 9, in <module>
    load_entry_point('pre-commit-hooks==0.6.0', 'console_scripts', 'trailing-whitespace-fixer')()
  File "/home/lucas_cimon/.pre-commit/repok3wjweh4/py_env-default/lib/python3.4/site-packages/pre_commit_hooks/trailing_whitespace_fixer.py", line 71, in fix_trailing_whitespace
    _fix_file(bad_whitespace_file, all_markdown or extension in md_exts)
  File "/home/lucas_cimon/.pre-commit/repok3wjweh4/py_env-default/lib/python3.4/site-packages/pre_commit_hooks/trailing_whitespace_fixer.py", line 12, in _fix_file
    for line in fileinput.input([filename], inplace=True):
  File "/usr/lib/python3.4/fileinput.py", line 263, in __next__
    line = self.readline()
  File "/usr/lib/python3.4/fileinput.py", line 363, in readline
    self._buffer = self._file.readlines(self._bufsize)
  File "/home/lucas_cimon/.pre-commit/repok3wjweh4/py_env-default/lib/python3.4/codecs.py", line 319, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 3: invalid continuation byte
```

And **all the content of `minimal_crasher.xml` lost** !!

In this test the xml file is very small, but it can happen with any file size nor number of staged modifications !

Lucas-C:


asottile:
Needs a test, also I'm not the biggest fan of fileinput, I think it might be easier just using `io.open` instead?  (that'd remove the awkward temporary file that may clobber other files in the working directory).

Lucas-C:
Test added

Lucas-C:
I'll do the move to `io.open` in a 2nd commit, when implementing the "option 2" we discussed.

asottile:
New test seems falling :(

Lucas-C:
Only the py34 AppVeyor test remains to fail.
I don't really get why actually

asottile:
When I'm looking for weird bugs that's music to my ears. 

I believe it's due to windows having a default encoding of cp1252

asottile:
Maybe it's time to approach the `'rb'` approach ;)

Lucas-C:
Done :)

Lucas-C:
AppVeyor is still failing :(
I wonder if it's because `readlines()` mangle newline bytes under Windows (it does not under Cygwin).
Sadly I'm going to be away from any computer for a week, so I won't make any progress on this in the following days.

asottile:
Ah yeah I've hit this before, the `newline=''` kwarg to open might fix this

asottile:
I'm off holiday in a few days so maybe I'll try and pick it up then

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -1,24 +1,28 @@
...
-        # preserve trailing two-space for non-blank lines in markdown files
-        if markdown and (not line.isspace()) and (line.endswith("  \n")):
-            line = line.rstrip(' \n')
-            # only preserve if there are no trailing tabs or unusual whitespace
-            if not line[-1].isspace():
-                print(line + "  ")
-                continue
-
-        print(line.rstrip())
+    with tempfile.NamedTemporaryFile(mode='w+b', delete=False) as tmp_file:
```
asottile: just store the old contents in memory imo

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -1,24 +1,28 @@
...
+        with open(filename, mode='rb') as original_file:
+            for line in original_file.readlines():
+                # preserve trailing two-space for non-blank lines in markdown files
+                eol = b'\r\n' if line[-2:] == b'\r\n' else b'\n'
+                if markdown and (not line.isspace()) and line.endswith(b'  ' + eol):
+                    line = line.rstrip(b' \r\n')  # restricted stripping: e.g. \t are not stripped
```
asottile: weird, I realize the lhs does this but I cannot remember why, should we also strip tabs?

pre_commit_hooks/trailing_whitespace_fixer.py
Lucas-C: Ok

pre_commit_hooks/trailing_whitespace_fixer.py
Lucas-C: If we replace this by a simple `rstrip()`, the following test fail:
```
    def test_markdown_linebreak_ext_opt(filename='bar.MD', input_s='bar   \nbaz\t   \n\t\n', output='bar  \nbaz\n\n')):
        ...
>       assert path.read() == output
E       assert 'bar  \nbaz  \n\n' == 'bar  \nbaz\n\n'
E           bar
E         - baz
E         ?    --
E         + baz
```

asottile: either way, I wouldn't worry about this if it means changing a test.  We can address that in a followup or not at all :)

Lucas-C: OK, fixes pushed

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -1,25 +1,29 @@
...
+        for line in lines:
+            file_processed.write(line)
 
+def _process_line(line, is_markdown):
+    # preserve trailing two-space for non-blank lines in markdown files
+    eol = b'\r\n' if line[-2:] == b'\r\n' else b'\n'
+    if is_markdown and (not line.isspace()) and line.endswith(b'  ' + eol):
+        line = line.rstrip()
+        # only preserve if there are no trailing tabs or unusual whitespace
+        if not line[-1:].isspace():
```
asottile: seems (from coverage) this is always true

asottile:
Stellar!  Thanks for the hard work on this :)

PR #142, opened 2016-09-01
end-of-file-fixer & trailing-whitespace: Add .scala and .ts files.

htmldoug:
We have some files that would like to benefit from all that whitespace fixing goodness.

@Lucas-C @asottile

### DIFF
--- a/hooks.yaml
+++ b/hooks.yaml
@@ -104,7 +104,7 @@
     description: Ensures that a file is either empty, or ends with one newline.
     entry: end-of-file-fixer
     language: python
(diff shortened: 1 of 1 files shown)


asottile:
Thanks!

Lucas-C:
I'm ok with this, but ideally we shouldn't have to include extra extensions by implementing https://github.com/pre-commit/pre-commit/pull/329

htmldoug:
Thanks for the quick reviews and merge.
https://github.com/pre-commit/pre-commit/pull/329 sounds like a nice addition.

PR #144, opened 2016-10-18
add j2 extension to trailing-whitespace and end-of-file-fixer

jzafran:
Add `.j2` file extension (Jinja2 templates) to trailing-whitespace and end-of-file-fixer


asottile:
Thanks!

PR #145, opened 2016-10-21
Use safe_load as per PyYAML documentation

### LINKED ISSUE #104: check-yaml should use `safe_load` instead of `load`

bdellegrazie:
resolves #104

### DIFF
--- a/pre_commit_hooks/check_yaml.py
+++ b/pre_commit_hooks/check_yaml.py
@@ -5,11 +5,10 @@ import sys
 
 import yaml
 
-
 try:
-    from yaml.cyaml import CLoader as Loader
+    from yaml.cyaml import CSafeLoader as Loader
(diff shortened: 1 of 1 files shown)


asottile:
Lg2m,  I'll verify the tests pass since Travis seems to be having issues and merge! 

Thanks!

bdellegrazie:
Hi @asottile -  I looked at the test output, turns out safe_load doesn't use a loader so the code is even simpler now.

asottile:
This will use the slower pure python implementation.  You can get the fast loader via `cyaml.CSafeLoader` and fall back to `yaml.SafeLoader` I believe

bdellegrazie:
Okay, will correct.

asottile:
You rock! Thanks for this :)

PR #147, opened 2016-11-03
adds top keys list of keys in hashes to put at the top of a hash

dmlb2000:
This adds custom sorting to preferentially add a list of top keys
at the start of any json hash in the json document

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -6,17 +6,95 @@ from collections import OrderedDict
 
 import simplejson
 
+class SortableOrderedDict(OrderedDict):
+    """Performs an in-place sort of the keys if you want."""
+    def sort(*args, **kwds):
+        self = args[0]
+        args = args[1:]
+        if 'key' not in kwds:
+            kwds['key'] = lambda x: x[0]
+        if len(args):
+            raise TypeError('expected no positional arguments got {0}'.format(len(args)))
+        sorted_od = sorted([x for x in self.items()], **kwds)
+        self.clear()
+        self.update(sorted_od)
 
-def _get_pretty_format(contents, indent, sort_keys=True):
-    return simplejson.dumps(
-        simplejson.loads(
-            contents,
-            object_pairs_hook=None if sort_keys else OrderedDict,
-        ),
-        sort_keys=sort_keys,
-        indent=indent
-    ) + "\n"  # dumps don't end with a newline
+class TrackedSod(SortableOrderedDict):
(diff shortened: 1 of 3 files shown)


dmlb2000:
I'd like some comments on how to test this and make the diff smaller.
The `KeyToCmp` class seems a bit over complicated and suggestions
on shrinking it would be welcome.

asottile:
So I think some of this can be simplified with functools.total_ordering, I also wonder if instead of keeping a global registry of objects whether the insertion / construction could establish the sorting.  I wouldn't worry about the performance here.

asottile:
Oh, as for tests, I'd just put some example inputs and outputs similar to how the other examples of this hook

dmlb2000:
@asottile 

The reason why I didn't do the sorting in the constructor was because I didn't know how simplejson would create the dictionary. If it creates an empty object then fills it with the individual key-values or if the constructor got the entirety of the key-value hash. Sorting afterwards seemed like a safer bet and tracking the objects as they get created by another code was pretty simple.

Currently, OrderedDict has an end-to-end connected linked list for keys as its data structure to handle consistent ordering. I think to make it work in `__setitem__` you'd have to copy everything out (like I'm doing in `sort()`) and put the data structures back in place sorted. The performance would suck but you wouldn't have to do tracking of all dicts at that point.

I'll definitely implement the total_ordering decorator that should reduce the need to have all the comparisons.

asottile:
The object pairs hook is just a callback which receives a list of pairs.  You could probably implement an efficient function which returns a normal OrderedDict :)

Something like:

```python
def pairs_first(pairs, first_pairs):
    before = [pair for pair in pairs if pair[0] in first_pairs]
    after = sorted([pair for pair in pairs if pair[0] not in first_pairs])
    return collections.OrderedDict(before + after)
```

and then `functools.partial` `first_pairs` and pass that as the `object_pairs_hook`

Something like:

```python
json.loads(s, object_pairs_hook=functools.partial(pairs_first, first_pairs=...))
```

dmlb2000:
@asottile Ah, I get it now... that's much cleaner... I'll get that fixed too.

karcaw:
Nice!

asottile:
Looks great!  Want to add a testing file which specifically targets your change?

dmlb2000:
I was actually able to do some of that with the existing test files, but I can add a more comprehensive example.

asottile:
Thanks! You rock

dmlb2000:
Bah! Just saw something, should have changed the README.md with the new argument... Pfft.

asottile:
Ah! Want to write up another quick PR to add some blurb to the README?

PR #149, opened 2016-11-12
Displaying the filename when the check-json hook fails with a UnicodeDecodeError - fix #148

### LINKED ISSUE #148: Could the check-json hook display the filename in case of a UnicodeDecodeError
Hi.

Waht do you think ? It is currently a bit of a pain to debug.

I can submit a PR.

Lucas-C:


asottile:
Thanks!

PR #151, opened 2016-11-30
Properly decode file contents.

ohlol:
Hi, this resolves an error using the detect_aws_credentials plugin on Python 3.

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -29,7 +29,7 @@ def check_file_for_aws_keys(filenames, keys):
 
     for filename in filenames:
         with open(filename, 'r') as content:
-            text_body = content.read()
+            text_body = content.read().decode('utf-8')
             if any(key in text_body for key in keys):
                 # naively match the entire file, low chance of incorrect collision
                 bad_files.append(filename)


asottile:
`open` in python3 returns a text stream so decode isn't correct (as you can see from the test output)

Is there a specific stacktrace you're seeing?

asottile:
my guess (without actually seeing anything) is you're hitting a case where `LANG=C` and it's using US-ASCII to read the file.

One way to force that (in a py2+py3 compatible way) would be to `import io` + `with io.open(..., encoding='UTF-8') as f:`

ohlol:
Ya I was getting this with the plugin enabled:

```
hookid: detect-aws-credentials

  File "/Users/scott/.pre-commit/repocg6re868/py_env-default/bin/detect-aws-credentials", line 11, in <module>
    load_entry_point('pre-commit-hooks==0.6.0', 'console_scripts', 'detect-aws-credentials')()
  File "/Users/scott/.pre-commit/repocg6re868/py_env-default/lib/python3.5/site-packages/pre_commit_hooks/detect_aws_credentials.py", line 62, in main
    bad_filenames = check_file_for_aws_keys(args.filenames, keys)
  File "/Users/scott/.pre-commit/repocg6re868/py_env-default/lib/python3.5/site-packages/pre_commit_hooks/detect_aws_credentials.py", line 32, in check_file_for_aws_keys
    text_body = content.read()
  File "/Users/scott/.pre-commit/repocg6re868/py_env-default/lib/python3.5/codecs.py", line 321, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcd in position 5: invalid continuation byte
```

ohlol:
Don't have LANG=C in my shell:

```% echo $LANG
en_US.UTF-8```

asottile:
That stacktrace indicates it is attempting to read a non-UTF8 file (most likely a binary file).  The hook is somewhat-naively (until we have a better solution) set up to check every file in your repository.  Ideally it would only check text files.  There's a WIP idea / implementation of this started: https://github.com/pre-commit/pre-commit/pull/329

I think a "fix" for this would be to make this hook ignore non-text files for now, I think we've done something similar for this hook: https://github.com/pre-commit/pre-commit-hooks/commit/4a01f64c8f67ec74edd2c6c501bce8bf61ad3494 (In this case, it stops reading the files as text and instead teaches them to read them as bytes.  For binary files, this'll succeed and won't attempt to decode them to text).

Maybe an approach like that would be the one to take here?  That or you can `exclude:` files which are binary manually, but that'd be a bit unfortunate I imagine :) 

Let me know if you'd like to take a stab at this!

asottile:
going to close this for now, it's not mergeable as-is

PR #152, opened 2016-11-30
Add a new hook to forbid new submodules

asottile:


chriskuehl:
This probably should've been added to `hooks.yaml`, huh.

PR #154, opened 2016-12-01
Fix test error "fatal: empty ident name (for <(null)>) not allowed"

alzeih:
This occurs when there is no global setting for git config options user.name and user.email

An example of the error shown below:
```
E           pre_commit.util.CalledProcessError: Command: ('/usr/bin/git', 'commit', '-m', 'init', '--allow-empty')
E           Return code: 128
E           Expected return code: 0
E           Output: (none)
E           Errors:
E
E               *** Please tell me who you are.
E
E               Run
E
E                 git config --global user.email "you@example.com"
E                 git config --global user.name "Your Name"
E
E               to set your account's default identity.
E               Omit --global to set the identity only in this repository.
E
E               fatal: empty ident name (for <(null)>) not allowed
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,7 @@ env: # These should match the tox env list
     - TOXENV=pypy
 install: pip install coveralls tox
 script: tox
-# Special snowflake.  Our tests depend on making real commits.
 before_install:
-  - git config --global user.name "Travis CI"
-  - git config --global user.email "user@example.com"
   # Install git-lfs for a test
   - './get-git-lfs.py && export PATH="/tmp/git-lfs:$PATH"'
 after_success:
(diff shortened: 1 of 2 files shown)


asottile:
This is already handled here: https://github.com/pre-commit/pre-commit-hooks/blob/77a7bba2f93cfe1205fc7167ef8379cbc6fcbb37/.travis.yml#L11-L13

(also you've got tabs)

alzeih:
@asottile It's not handled when running the test from tox locally though

I'll fix the tabs in a rebase :/

asottile:
Thanks for the PR

asottile:
I imagine pre-commit/pre-commit  could benefit from the same change

Issue #157, opened 2016-12-07
[Question] why can't we add pylint in this repo hooks.yaml file?


lavish205:
My `.pre-commit-config.yaml` file looks like this:
```yaml
-   repo: https://github.com/pre-commit/pre-commit-hooks
     sha: 52582865ab7ed8f41734253190844c854ddd9a9e
     hooks:
     -   id: trailing-whitespace
     -   id: check-ast
     -   id: debug-statements
     -   id: end-of-file-fixer
-   repo: git://github.com/pre-commit/mirrors-pylint
     sha: v1.6.4
     hooks:
    -   id: pylint
```
and when I run `pre-commit install` it creates a repo for pylint in `~/.pre-commit/repoXXXX/` with its own virtual env in it, due to which pylint shows `import errors`.
So there are two ways to solve this error:

1. add all required packages in `additional_dependencies` in `.pre-commit-config.yaml` file for pylint.
But I can't add all the packages here because it useless to create two virtual env with same packages.
Can we however, merge it with our own virtualenv??

2. if I modify `pre-commit-hooks/hooks.yaml` and add pylint hook over there, it works perfectly fine.

Can you suggest me which is the best way to do it??

asottile:
So tl;dr is pylint does some static analysis and some dynamic analysis.  When put in an isolated virtualenv it does not work properly.  Using isolated virtualenvs achieves the design goal of reproducibility in pre-commit.

That said, there are still a few options for using pylint with pre-commit:

- disable the dynamic checking parts (not usually the desirable option)
- use it as a "system" hook (usually through "local" hooks).

You can find some related issues:
- https://github.com/pre-commit/pre-commit/issues/362
- https://github.com/pre-commit/pre-commit/issues/266
- https://github.com/pre-commit/pre-commit/issues/178
- https://github.com/Yelp/venv-update/blob/de9e1befdecf448cf02ceb5c57a30bcc88a58c06/.pre-commit-config.yaml#L30-L36
- http://pre-commit.com/#repository-local-hooks

swateek:
[https://github.com/Yelp/venv-update/blob/de9e1befdecf448cf02ceb5c57a30bcc88a58c06/.pre-commit-config.yaml#L30-L36](https://github.com/Yelp/venv-update/blob/de9e1befdecf448cf02ceb5c57a30bcc88a58c06/.pre-commit-config.yaml#L30-L36)

is a good example, thank you!

Also, needed to install pylint

[...]

PR #158, opened 2016-12-07
fixes issue #157: adding pylint hook support

lavish205:
#157


asottile:
For pylint, I would suggest either using https://github.com/pre-commit/mirrors-pylint with dynamic analysis disabled or use it as a repository local hook http://pre-commit.com/#repository-local-hooks (for example https://github.com/Yelp/venv-update/blob/de9e1befdecf448cf02ceb5c57a30bcc88a58c06/.pre-commit-config.yaml#L30-L36)

PR #160, opened 2016-12-26
Detect Ed25519 keys as well

Dunedan:
`Ed255519` keys generated by OpenSSH contain "`BEGIN OPENSSH PRIVATE KEY`"
as identifier. This commit adds coverage for such keys as well.


chriskuehl:
Thanks!

PR #162, opened 2016-12-30
Improve searching for configured AWS credentials

Dunedan:
The previous approach for finding AWS credentials was pretty naive and
only covered contents of a single file (`~/.aws/credentials` by default).

The AWS CLI documentation states various other ways to configure
credentials which weren't covered:
https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#credentials
Even that aren't all ways, a look into the code shows:
https://github.com/boto/botocore/blob/develop/botocore/credentials.py

This commit changes the behavior so the hook will behave in a way
that if the AWS CLI is able to obtain credentials from local files,
the hook will find them as well.

The changes in detail are:
- detect AWS session tokens and handle them like secret keys.
- always search credentials in the default AWS CLI file locations
  (`~/.aws/config`, `~/.aws/credentials`, `/etc/boto.cfg` and `~/.boto`)
- detect AWS credentials configured via environment variables in
  `AWS_SECRET_ACCESS_KEY`, `AWS_SECURITY_TOKEN` and `AWS_SESSION_TOKEN`
- check additional configuration files configured via environment
  variables (`AWS_CREDENTIAL_FILE`, `AWS_SHARED_CREDENTIALS_FILE` and
  `BOTO_CONFIG`)
- print out the first four characters of each secret found in files to
  be checked in, to make it easier to figure out, what the secrets
  were, which were going to be checked in
- improve error handling for parsing ini-files
- improve tests

There is a major functional change introduced by this commit:
Locations the AWS CLI gets credentials from are always searched and
there is no way to disable them. `--credentials-file` is still there to
specify one or more additional files to search credentials in. It's
the purpose of this hook to find and check files for found
credentials, so it should work in any case. As this commit also
improves error handling for not-existing or malformed configuration
files, it should be no big deal.

Receiving credentials via the EC2 and ECS meta data services is not
covered intentionally, to not further increase the amount of changes
in this commit and as it's probably an edge case anyway to have this
hook running in such an environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,12 @@ Add this to your `.pre-commit-config.yaml`
 - `check-xml` - Attempts to load all xml files to verify syntax.
 - `check-yaml` - Attempts to load all yaml files to verify syntax.
 - `debug-statements` - Check for pdb / ipdb / pudb statements in code.
-- `detect-aws-credentials` - Checks for the existence of AWS secrets that you have set up with the AWS CLI.
+- `detect-aws-credentials` - Checks for the existence of AWS secrets that you
+  have set up with the AWS CLI.
+  The following arguments are available:
+  - `--credential-file` - additional AWS CLI style configuration file in a
+    non-standard location to fetch configured credentials from. Can be repeated
+    multiple times.
 - `detect-private-key` - Checks for the existence of private keys.
 - `double-quote-string-fixer` - This hook replaces double quoted strings
   with single quoted strings.
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Add this to your `.pre-commit-config.yaml`
 - `detect-aws-credentials` - Checks for the existence of AWS secrets that you
   have set up with the AWS CLI.
   The following arguments are available:
-  - `--credential-file` - additional AWS CLI style configuration file in a
+  - `--credentials-file` - additional AWS CLI style configuration file in a
     non-standard location to fetch configured credentials from. Can be repeated
     multiple times.
 - `detect-private-key` - Checks for the existence of private keys.
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -10,24 +10,23 @@ from six.moves import configparser
 def get_aws_credential_files_from_env():
     """Extract credential file paths from environment variables."""
     files = set()
-    for env_var in {'AWS_CONFIG_FILE', 'AWS_CREDENTIAL_FILE',
-                    'AWS_SHARED_CREDENTIALS_FILE', 'BOTO_CONFIG'}:
-        try:
+    for env_var in (
+        'AWS_CONFIG_FILE', 'AWS_CREDENTIAL_FILE', 'AWS_SHARED_CREDENTIALS_FILE',
+        'BOTO_CONFIG'
+    ):
+        if env_var in os.environ:
             files.add(os.environ[env_var])
-        except KeyError:
-            pass
     return files
 
 
 def get_aws_secrets_from_env():
     """Extract AWS secrets from environment variables."""
     keys = set()
-    for env_var in {'AWS_SECRET_ACCESS_KEY', 'AWS_SECURITY_TOKEN',
-                    'AWS_SESSION_TOKEN'}:
-        try:
+    for env_var in (
+        'AWS_SECRET_ACCESS_KEY', 'AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN'
+    ):
+        if env_var in os.environ:
             keys.add(os.environ[env_var])
-        except KeyError:
-            pass
     return keys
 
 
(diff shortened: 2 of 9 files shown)


arahayrabedian:
Original hook author here, I like what you've done with the place :+1: (don't take this as review though)

README.md
```diff
@@ -40,7 +40,12 @@ Add this to your `.pre-commit-config.yaml`
 - `check-xml` - Attempts to load all xml files to verify syntax.
 - `check-yaml` - Attempts to load all yaml files to verify syntax.
 - `debug-statements` - Check for pdb / ipdb / pudb statements in code.
-- `detect-aws-credentials` - Checks for the existence of AWS secrets that you have set up with the AWS CLI.
+- `detect-aws-credentials` - Checks for the existence of AWS secrets that you
+  have set up with the AWS CLI.
+  The following arguments are available:
+  - `--credential-file` - additional AWS CLI style configuration file in a
```
arahayrabedian: typo --> should be `--credentials-file`, not usually a stickler but this is a readme.

Dunedan: Thanks for pointing that out. I fixed it with a recent commit.
I wish there'd be a hook to prevent checking in such typos. :wink:

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -7,62 +7,118 @@
...
 
 
-def get_your_keys(credentials_file):
-    """reads the secret keys in your credentials file in order to be able to
-    look for them in the submitted code.
+def get_aws_credential_files_from_env():
+    """Extract credential file paths from environment variables."""
+    files = set()
+    for env_var in {'AWS_CONFIG_FILE', 'AWS_CREDENTIAL_FILE',
+                    'AWS_SHARED_CREDENTIALS_FILE', 'BOTO_CONFIG'}:
```
asottile: p4: I'd iterate over a tuple instead of a set, you don't need set semantics here (same applies for the function below)

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -7,62 +7,118 @@
...
+        try:
+            files.add(os.environ[env_var])
+        except KeyError:
```
asottile: p4: it's fewer lines and slightly more straghtforward to write:

```python
if env_var in os.environ:
    files.add(os.environ[env_var])
```

(same applies for the function below)

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -7,62 +7,118 @@
...
-            if any(key in text_body for key in keys):
-                # naively match the entire file, low chance of incorrect collision
-                bad_files.append(filename)
-
+            for key in keys:
+                # naively match the entire file, low chance of incorrect
+                # collision
+                if key in text_body:
+                    bad_files.append({'filename': filename,
+                                      'key': key[:4].ljust(32, str('*'))})
```
asottile: p4: without needing ljust: `key[:4] + '*' * 28`

tests/detect_aws_credentials_test.py
```diff
@@ -1,38 +1,103 @@
...
+    monkeypatch.setenv('AWS_CREDENTIAL_FILE', '/foo')
+    assert get_aws_credential_files_from_env() == {'/foo'}
+    monkeypatch.setenv('AWS_SHARED_CREDENTIALS_FILE', '/bar')
+    assert get_aws_credential_files_from_env() == {'/foo', '/bar'}
+    monkeypatch.setenv('BOTO_CONFIG', '/baz')
+    assert get_aws_credential_files_from_env() == {'/foo', '/bar', '/baz'}
+    monkeypatch.setenv('AWS_CONFIG_FILE', '/xxx')
+    assert get_aws_credential_files_from_env() == {'/foo', '/bar', '/baz', '/xxx'}
+    monkeypatch.setenv('AWS_DUMMY_KEY', 'foobar')
+    assert get_aws_credential_files_from_env() == {'/foo', '/bar', '/baz', '/xxx'}
```
asottile: to me this feels like 5-6 different tests.  You may find it easier to write this test with `mock.patch.dict` as such (and I would suggest it since it definitely improves readability):


```python
@pytest.mark.parametrize(
    ('environ', 'expected'),
    (
        ({'AWS_CREDENTIAL_FILE': '/foo'}, {'/foo'}),
        # etc. ...
    ),
)
def test_get_aws_credentials_file_from_env(environ, expected):
    with mock.patch.dict(os.environ, environ, clear=True):
        # ...
```

I also tend to steer away from the `monkeypatch` fixture due to not being able to tightly control when the fixture gets applied / unapplied and prefering explicit use of `mock.patch`, but I can let that go if it's too much of a nitpick :)

Dunedan:
Definitely not too much nitpicking. I highly appreciate the feedback. I'll soon push another commit, fixing the mentioned issues.

asottile:
Looks like appveyor is only grumpy because of this: https://github.com/pre-commit/pre-commit-hooks/pull/163

unclear why it didn't trigger in travis, but shrugs!

asottile:
Thanks for the PR!

PR #164, opened 2017-01-19
No commit branch hook addition.

karcaw:


karcaw:
Did not mean for this to come to upstream yet..

PR #170, opened 2017-01-31
Pretty yaml

macisamuele:
As exists ``pretty-format-json`` I have added an equivalent hook for yaml files.
Since formatting of yaml files is heavily dependent on parameters passed to ``yaml.dump`` method I have decided to expose all the yaml.dump ``kwargs``  parameters in order to let the users of the hook to decide how a pretty YAML file looks like!

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -49,6 +49,12 @@
     entry: pretty-format-json
     language: python
     files: \.json$
+-   id: pretty-format-yaml
+    name: Pretty format YAML
+    description: This hook sets a standard for formatting YAML files.
+    entry: pretty-format-yaml
+    language: python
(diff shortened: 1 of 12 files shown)


asottile:
yaml isn't round trippable so this hook isn't valid -- at best this will work for things which have no comments and no human decided formatting.  At worst this will be a constant frustration for anyone trying to add comments / lists formatting / refs / any of the other features of yaml in the 72 page spec.

You can see (struggled) attempts to implement something like this: pre-commit/pre-commit#211 pre-commit/pre-commit#415

macisamuele:
@asottile++ I wasn't thinking about comments ... I'm closing the PR!

asottile:
<3 we've tried making this, and ran into the same things (why it doesn't exist!) :D

PR #175, opened 2017-02-10
Add flag to detect-aws-credentials to allow missing keys

### LINKED ISSUE #174: detect-aws-credentials fails when no credentials are set
As part of our developer setup, each person installs the pre-commit and inherits the project's hooks, but not everyone has aws credentials set up, so `detect-aws-credentials` will product a failure when the individual hasn't set any keys file or has any exported to their env.

https://github.com/pre-commit/pre-commit-hooks/blob/20f04626a1ee7eb34955d775a37aa9a56a0a7448/pre_commit_hooks/detect_aws_credentials.py#L112

I'm not sure of the best way to handle this within the current logic - I'd rather exit 0 with a passing warning than fail, since this then forces the individual to go and set up a credentials file, even if they don't use one.

I guess the "simple" workaround is to `export AWS_SECRET_ACCESS_KEY="RANDOMSTRINGTHATNEVERTEXISTSINMYCODE!!"` before running the pre-commit hooks, but this seems wrong.

miketheman:
In the event that there exists no configuration for AWS credentials and
they are not exported in to the current environment, a flag may be
passed to allow exiting the hook successfully.

Resolves #174

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -95,6 +95,12 @@ def main(argv=None):
             'secret keys from'
         )
     )
+    parser.add_argument(
+        '--allow-missing-credentials',
+        dest='allow_missing_credentials',
+        action='store_true',
+        help='Allow hook to pass when no credentials are detected.'
+    )
     args = parser.parse_args(argv)
 
     credential_files = set(args.credential_files)
@@ -110,6 +116,9 @@ def main(argv=None):
     # Secrets might be part of environment variables, so add such secrets to
     # the set of keys.
     keys |= get_aws_secrets_from_env()
+
+    if not keys and args.allow_missing_credentials:
+        return 0
 
     if not keys:
         print(
--- a/tests/detect_aws_credentials_test.py
+++ b/tests/detect_aws_credentials_test.py
@@ -130,3 +130,17 @@ def test_non_existent_credentials(mock_secrets_env, mock_secrets_file, capsys):
         'and environment variables.\nPlease ensure you have the '
         'correct setting for --credentials-file\n'
     )
+
+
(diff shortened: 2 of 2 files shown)


CHANGELOG
```diff
@@ -1,3 +1,7 @@
+0.8.0
+=====
+- Add flag to allow aws detection to pass when no keys exist
+
```
asottile: Don't bother with the changelog, I'll fill this out when I decide to make a release <3

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -95,6 +95,12 @@ def main(argv=None):
             'secret keys from'
         )
     )
+    parser.add_argument(
+        '--allow-missing-credentials',
+        dest='allow_missing_credentials',
+        action='store_true',
+        help=('Allow hook to pass when no credentials are detected.')
```
asottile: parens aren't necessary here

miketheman:
Comments addressed, thanks!

asottile:
Thanks!

AdityaSoni19031997:
Am I doing something incorrect wrt `detect-aws-credentials`?

```yaml
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.3.0
    hooks:
      - id: detect-private-key
      - id: detect-aws-credentials
        args: ["--allow-missing-credentials"]
```

I tried passing the flag but still it fails at the `aws-creds` hook?

```
No AWS keys were found in the configured credential files and environment variables. Please ensure you have the correct setting for --credentials-file.
```

asottile:
don't edit it, you've edited the output

AdityaSoni19031997:
Nevermind @asottile. It was some intermittent issue, we're good here and everything is working as expected!

Apologies for the false alarm.

Thanks

PR #177, opened 2017-02-25
Change `sha` in readme to be a real value

asottile:
I noticed in our traffic on pre-commit/pre-commit that there were a lot of hits to [this issue](https://github.com/pre-commit/pre-commit/issues/366) presumably due to copy pasting from this readme.

This should reduce frustration in that regard (we'll have to remember to update this whenever bumping the version).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ See also: https://github.com/pre-commit/pre-commit
 Add this to your `.pre-commit-config.yaml`
 
     -   repo: git://github.com/pre-commit/pre-commit-hooks
-        sha: ''  # Use the sha you want to point at
+        sha: v0.7.1  # Use the ref you want to point at
         hooks:
         -   id: trailing-whitespace
         # -   id: ...


arahayrabedian:
@asottile any reason not to write `master` and not worry about it?

arahayrabedian:
3 minutes of reading up sorted that out -_-

asottile:
Reproducibility :)  You can read more about the reasoning here: https://github.com/pre-commit/pre-commit/issues/158#issuecomment-54103765

asottile:
Actually, if you google any part of that error message (even without the exact error message) you end up with our issue in the top 5 results.  gg google.  Either way I think this is a useful improvement.

chriskuehl:
I wonder how hard it would be to get http://pre-commit.com/hooks.html to show the latest tag on each repo.

asottile:
It should be pretty easy, we're already cloning each repo to make that page: https://github.com/pre-commit/pre-commit.github.io/blob/23407079ee24b2df514cf409af3b197d7ea1b76a/make_all_hooks.py#L14-L26

Issue #181, opened 2017-03-15
django tests.py is not handled properly


oliver-zhou:
Issue in #88 introduced a minor bug.

Django creates "tests.py" as the default test file name when you run `./manage.py startapp` and now pre-commit-hooks doesn't handle it properly, seeking only files with test_*.py

Since the last fix added os.path.basename, we're always working with the filename, and should return to using test.*.py

Additionally - readme.md still references test*.py as well, and not test_*.py

The ultra-best-right way to handle this would be to handle it how python does it. unittest.TestLoader.discover does it by allowing any pattern, defaulting to test.py when no argument is provided.

https://github.com/python/cpython/blob/master/Lib/unittest/loader.py 
    def discover(self, start_dir, pattern='test*.py', top_level_dir=None):

asottile:
via #182

PR #182, opened 2017-03-15
Handling default django test filename of "tests.py"

oliver-zhou:
#181 - Fix to restore regex to handle tests.py, the default Django test format


asottile:
Looks good!

From https://docs.djangoproject.com/en/1.10/topics/testing/overview/

> Test discovery is based on the unittest module’s built-in test discovery. By default, this will discover tests in any file named “test*.py” under the current working directory.

asottile:
Thanks for the PR :D

PR #184, opened 2017-03-16
Add --no-ensure-ascii option to pretty-format-json hook

ushuz:
For CJK users like me, it makes JSON files more legible in text editor if non-ASCII characters can be kept as-is without being converted into Unicode escape sequences. For example, `["中文"]` is far more better than `["\u4e2d\u6587"]`.

But it will fail the check due to `contents != pretty_contents`. `["中文"]` is acutually `["\xe4\xb8\xad\xe6\x96\x87"]` in a UTF-8 file, so is `contents`, a bytes string with non-ASCII characters. While `pretty_contents` is the return value of `simplejson.dumps()`, under default configuration (`ensure_ascii=True`), a bytes string with Unicode escape sequences.

In plain Python,

```python
contents == '["\xe4\xb8\xad\xe6\x96\x87"]'
pretty_contents == '["\\u4e2d\\u6587"]'
```

Here's the fix:
1. Make sure `contents` and `pretty_contents` are both `unicode`, so they can be compared properly
2. Offer an option to set `ensure_ascii` to `False`

Then,

```python
contents == pretty_contents == u'["\u4e2d\u6587"]'
```


tests/pretty_format_json_test.py
```diff
@@ -48,6 +51,17 @@ def test_tab_pretty_format_json(filename, expected_retval):
     assert ret == expected_retval
 
 
+@pytest.mark.parametrize(('filename', 'expected_retval'), (
+    ('not_pretty_formatted_json.json', 1),
+    ('unsorted_pretty_formatted_json.json', 1),
+    ('non_ascii_pretty_formatted_json.json', 0),
+    ('pretty_formatted_json.json', 0),
+))
+def test_non_ascii_pretty_format_json(filename, expected_retval):
```
asottile: I wouldn't bother to parametrize this, you're really only trying to demonstrate the one file :shrugs:

asottile:
Seems windows is failing with this change

tests/pretty_format_json_test.py
ushuz: Addressed

ushuz:
@asottile It's quite tricky on Windows. `check-json` hook can't properly decode `non_ascii_pretty_formatted_json.json` on Windows with Python 3.6. The file is UTF-8 encoded, but `open()` in Python 3.x will decode input file automatically using `locale.getpreferredencoding()`.

pre_commit_hooks/pretty_format_json.py
```diff
@@ -90,20 +100,23 @@ def pretty_format_json(argv=None):
     status = 0
 
     for json_file in args.filenames:
-        with open(json_file) as f:
+        with io.open(json_file, encoding='utf-8') as f:
```
ushuz: @asottile I'm assuming all inputs are UTF-8 compatible here. Not sure this is a good way to handle it.

asottile: I think that's a reasonable assumption, the json spec indicates they should have some UTF-x encoding

asottile:
Thanks!

PR #185, opened 2017-03-17
Add a no commit to specific branch hook.

karcaw:


pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -0,0 +1,32 @@
...
+from pre_commit_hooks.util import cmd_output
+
+
+def is_on_branch(protected):
+    retval = False
+    branch = cmd_output('git', 'symbolic-ref', 'HEAD')
+    chunks = branch.strip().split('/')
+    if chunks[2] == protected:
+        retval = True
+    return retval
```
asottile: this `if ...: return True return False` can just be `return chunks[2] == protected`

pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-b', default='master', help='branch to disallow commits to')
+    parser.add_argument('filenames', nargs='*', help='filenames to check.')
```
asottile: note for `always_run` hooks, they will receive no arguments

pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -0,0 +1,32 @@
...
+    args = parser.parse_args(argv)
+
+    if is_on_branch(args.b):
+        return 1
+    else:
+        return 0
```
asottile: could be simplified to `return int(is_on_branch(args.b))`

pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -0,0 +1,32 @@
...
+    chunks = branch.strip().split('/')
+    if chunks[2] == protected:
+        retval = True
+    return retval
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-b', default='master', help='branch to disallow commits to')
```
asottile: maybe add a longopt `--branch` so the code can read easier `args.branch`

setup.py
```diff
@@ -55,6 +55,7 @@
             'pretty-format-json = pre_commit_hooks.pretty_format_json:pretty_format_json',
             'requirements-txt-fixer = pre_commit_hooks.requirements_txt_fixer:fix_requirements_txt',
             'trailing-whitespace-fixer = pre_commit_hooks.trailing_whitespace_fixer:fix_trailing_whitespace',
+            'no-commit-to-branch = pre_commit_hooks.no_commit_to_branch:main',
```
asottile: can you alphabetize this?

pre_commit_hooks/no_commit_to_branch.py
karcaw: I like that better too

pre_commit_hooks/no_commit_to_branch.py
karcaw: removing..

pre_commit_hooks/no_commit_to_branch.py
karcaw: ok!

pre_commit_hooks/no_commit_to_branch.py
karcaw: removes a few lines of code too.

setup.py
karcaw: yes

.pre-commit-hooks.yaml
```diff
@@ -147,3 +147,9 @@
     entry: trailing-whitespace-fixer
     language: python
     files: \.(asciidoc|adoc|coffee|cpp|css|c|ejs|erb|groovy|h|haml|hh|hpp|hxx|html|in|j2|jade|json|js|less|markdown|md|ml|mli|pp|py|rb|rs|R|scala|scss|sh|slim|tex|tmpl|ts|txt|yaml|yml)$
+-   id: no-commit-to-branch
+    name: Dont commit to branch
```
chriskuehl: nit: apostrophe in "Don't"? You might have to fool YAML by not using a bareword, e.g.

```yaml
name: "Don't commit to branch"
```

karcaw: that seemed to work, I went ahead and put the hook in alphabetic order too.

pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -0,0 +1,25 @@
...
+import argparse
+import sys
+
+from pre_commit_hooks.util import cmd_output
+
+
+def is_on_branch(protected):
+    branch = cmd_output('git', 'symbolic-ref', 'HEAD')
+    chunks = branch.strip().split('/')
+    return chunks[2] == protected
```
asottile: oops sorry I didn't catch this at first, I think this should be:

```python
return '/'.join(chunks[2:]) == protected
```

branches can have slashes in them (maybe add a test for this too?)

karcaw: Adding Tests for this change.

asottile:
Thanks for the PR!

I'm currently fixing the windows tests, they're grumpy about the newly added non-utf8 file

PR #189, opened 2017-03-27
no-commit-to-branch: Fixed manifest validation issue

andir:
Since https://github.com/pre-commit/pre-commit has changed to it's now schema validation the no-commit-to-branch check failed. The `files` attribute should not be present when `always_run` is `true`

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -133,7 +133,6 @@
     name: "Don't commit to branch"
     entry: no-commit-to-branch
     language: python
-    files: .*
     always_run: true
 -   id: pyflakes
     name: Pyflakes (DEPRECATED, use flake8)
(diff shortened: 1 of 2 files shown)


andir:
This is a nice recursive issue... the CI pulls the hooks (`https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml#L1`) from an older commit which of course does not work with newer `pre-commit` versions...

asottile:
Woops, this was a mistake made in pre-commit/pre-commit#502, I'll be re-releasing a new version without this problem

asottile:
I've uploaded 0.13.6 -- to ensure backward compatibility (which I mistakenly broke oops) I think we'll want to wait on this change until there's wider adoption of the latest release.  Thanks for the PR though :)

andir:
Np, thanks for fixing it. My search voodoo was weak :-)

PR #190, opened 2017-04-26
define more branch protected

moas:


asottile:
This fails the tests as written -- want to fix those up?

asottile:
via #294

PR #200, opened 2017-06-13
demo

tom-aa:


asottile:
I assume this was a mistake?  btw you can make `pep8` work just by doing the following:

```yaml
-   id: pep8
    entry: pep8
    files: '\.py$'
    language: python
```

and the framework will take care of passing the correct files, etc. to your hook

(though pep8 has been renamed to pycodestyle)

PR #201, opened 2017-06-19
edit detect_private_key to ignore directories

jakob223:
Currently if given a directory, detect_private_key tries to open it as a file and spews errors as a result. The behavior of this should either recursively search the directory or ignore the directory instead. This code ignores the directory.

### DIFF
--- a/pre_commit_hooks/detect_private_key.py
+++ b/pre_commit_hooks/detect_private_key.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import argparse
+import os
 import sys
 
 BLACKLIST = [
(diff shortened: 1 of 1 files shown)


asottile:
`pre-commit` should never pass a directory to this -- I'm not sure this is a case that should be defended against?

jakob223:
Hmm, I had a check failing because it was trying to open directories. I'll check back tomorrow and update with more information.

asottile:
Ah yeah, probably this issue: https://github.com/pre-commit/pre-commit/issues/191#issuecomment-113249998 (running against submodules)

we plan to handle this in a more blanket way in the framework such that every hook doesn't need to guard against running on directories.

There's a wip branch which enables hooks "by type" (which would disable linting "subodule directories" by default unless a hook opted into them).  @chriskuehl and I hope to finalize this branch _soon_ (though we've both been pretty busy)

jakob223:
Yep, now I remember that it was a submodule. If this is being solved in that way, then I'm happy to close this PR.

asottile:
This is now the default behaviour (see the linked ticket) in 0.15.0.  See pre-commit/pre-commit#551

PR #202, opened 2017-06-23
File contents sorter hook

dlgallagher:
(this is my first github pull request, so please let me know if there's some etiquette or anything I should know about)

This branch adds a new precommit hook file-contents-sorter that does just what it sounds like: sorts the lines in files provided to it as arguments. It's not meant to run on any files by default; it must have filenames provided to it by the user in .pre-commit-config.yaml.

I want this for sorting one-thing-per-line kind of files in our services like deploy-whitelist.txt or .gitignore.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -110,7 +110,7 @@
     description: Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input in your .pre-commit-config.yaml file.
     entry: file-contents-sorter
     language: python
-    files: ''
+    files: '^$'
 -   id: fix-encoding-pragma
     name: Fix python encoding pragma
     language: python
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -105,6 +105,12 @@
     entry: end-of-file-fixer
     language: python
     files: \.(asciidoc|adoc|coffee|cpp|css|c|ejs|erb|groovy|h|haml|hh|hpp|hxx|html|in|j2|jade|json|js|less|markdown|md|ml|mli|pp|py|rb|rs|R|scala|scss|sh|slim|tex|tmpl|ts|txt|yaml|yml)$
+-   id: file-contents-sorter
+    name: File Contents Sorter
+    description: Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input in your .pre-commit-config.yaml file.
+    entry: file-contents-sorter
+    language: python
+    files: ''
 -   id: fix-encoding-pragma
     name: Fix python encoding pragma
     language: python
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -153,6 +153,12 @@
     entry: requirements-txt-fixer
     language: python
     files: requirements.*\.txt$
+-   id: sort-simple-yaml
+    name: Sort simple YAML files
+    language: python
+    entry: sort-simple-yaml
+    description: Sorts simple YAML files which consist only of top-level keys, preserving comments and blocks.
+    files: '^$'
 -   id: trailing-whitespace
     name: Trim Trailing Whitespace
     description: This hook trims trailing whitespace.
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -1,0 +1,57 @@
+"""
+A very simple pre-commit hook that, when passed one or more filenames
+as arguments, will sort the lines in those files.
+
+An example use case for this: you have a deploy-whitelist.txt file
+in a repo that contains a list of filenames that is used to specify
+files to be included in a docker container. This file has one filename
+per line. Various users are adding/removing lines from this file; using
+this hook on that file should reduce the instances of git merge
+conflicts and keep the file nicely ordered.
+"""
+from __future__ import print_function
+
+import argparse
+
+PASS = 0
+FAIL = 1
+
+
+def sort_file_contents(f):
+    before = [line for line in f]
+    after = sorted(before)
+
+    before_string = b''.join(before)
+    after_string = b''.join(after)
+
+    if before_string == after_string:
+        return PASS
+    else:
+        f.seek(0)
+        f.write(after_string)
+        f.truncate()
+        return FAIL
+
+
+def parse_commandline_input(argv):
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='+', help='Files to sort')
+    args = parser.parse_args(argv)
+    return args
+
+
+def main(argv=None):
+    args = parse_commandline_input(argv)
+
+    retv = PASS
+
+    for arg in args.filenames:
+        with open(arg, 'rb+') as file_obj:
+            ret_for_file = sort_file_contents(file_obj)
+
+            if ret_for_file:
+                print('Sorting {}'.format(arg))
+
+            retv |= ret_for_file
+
+    return retv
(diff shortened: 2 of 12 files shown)


.pre-commit-hooks.yaml
```diff
@@ -105,6 +105,12 @@
     entry: end-of-file-fixer
     language: python
     files: \.(asciidoc|adoc|coffee|cpp|css|c|ejs|erb|groovy|h|haml|hh|hpp|hxx|html|in|j2|jade|json|js|less|markdown|md|ml|mli|pp|py|rb|rs|R|scala|scss|sh|slim|tex|tmpl|ts|txt|yaml|yml)$
+-   id: file-contents-sorter
+    name: File Contents Sorter
+    description: Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input in your .pre-commit-config.yaml file.
+    entry: file-contents-sorter
+    language: python
+    files: ''
```
asottile: this by default matches *all* files, to match none I'd suggest `^$`

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -0,0 +1,57 @@
...
+from __future__ import print_function
+
+import argparse
+
+PASS = 0
+FAIL = 1
+
+
+def sort_file_contents(f):
+    before = [line for line in f]
```
asottile: aka `list(f)`

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -0,0 +1,57 @@
...
+        f.write(after_string)
+        f.truncate()
+        return FAIL
+
+
+def parse_commandline_input(argv):
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='+', help='Files to sort')
+    args = parser.parse_args(argv)
+    return args
```
asottile: just put this in `main`

requirements-dev.txt
```diff
@@ -2,6 +2,7 @@
 
 coverage
 flake8
+ipdb
```
asottile: please don't add ipdb (rebase out) -- if you want this in your own environment install it there

tests/file_contents_sorter_test.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+
+from pre_commit_hooks.file_contents_sorter import FAIL
+from pre_commit_hooks.file_contents_sorter import main
+from pre_commit_hooks.file_contents_sorter import parse_commandline_input
+from pre_commit_hooks.file_contents_sorter import PASS
+
+
+def _n(*strs):
+    return b'\n'.join(strs) + b'\n'
```
asottile: `_n` to me feels like "native string" -- I'd also just write the string literals directly -- generally [avoid logic in tests](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html)

tests/file_contents_sorter_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.mark.parametrize(
+    ('filename_list'),
+    (
+        ['filename1'],
+        ['filename1', 'filename2'],
+    )
+)
+def test_parse_commandline_input_success(filename_list):
```
asottile: I wouldn't bother writing a test for this, you're really just testing if argparse does what argparse does -- I think you can assume `+` works as written on the tin

asottile:
woops, github posted before I finished -- there's also a closed source hook in `yelp_pre_commit_hooks` (if I remember correctly) that @chriskuehl wrote that sorts *simply* yaml nicely -- that might be worth open sourcing too :)

asottile:
You can also ignore the appveyor failure -- it's due to a breaking change in `git-lfs` (of which I'm waiting for a new release [here](https://github.com/git-lfs/git-lfs/pull/2311)) -- I've restarted travis-ci (the caches were invalid)

pre_commit_hooks/file_contents_sorter.py
dlgallagher: I saw that most other hooks have the args parsed in their main function. Reasons I made it a separate function:
1) I found it made main() method more readable
2) unit testing

There's probably some I've-been-brainwashed-by-yelp_batch-which-does-this going on too.

Since the file is so short anyway, do you mind if I just leave it as a separate fn?

.pre-commit-hooks.yaml
dlgallagher: Thanks! I'll update that.

requirements-dev.txt
dlgallagher: OK I'll remove it. 
This sounds like one of those etiquette things... is adding a new dev dependency here a general no-no? Or is there another reason for it?

tests/file_contents_sorter_test.py
dlgallagher: OK will update that.

tests/file_contents_sorter_test.py
dlgallagher: OK; removed. I think I'd been worried about protecting against future changes to the parsing breaking things. The code is so simple though maybe just not worth worrying about.

pre_commit_hooks/file_contents_sorter.py
asottile: None of the other hooks in this repository do this -- yelp_batch has a reason to do this (it is **unnecessarily** complicated).

tests/file_contents_sorter_test.py
```diff
@@ -0,0 +1,30 @@
...
+    (b'alpha\nbeta\n', PASS, b'alpha\nbeta\n'),
+    (b'beta\nalpha\n', FAIL, b'alpha\nbeta\n'),
+    (b'C\nc\n', PASS, b'C\nc\n'),
+    (b'c\nC\n', FAIL, b'C\nc\n'),
+    (b'mag ical \n tre vor\n', FAIL, b' tre vor\nmag ical \n'),
+    (b'@\n-\n_\n#\n', FAIL, b'#\n-\n@\n_\n'),
+)
+
+
+@pytest.mark.parametrize(('input_s', 'expected_retval', 'output'), TESTS)
```
asottile: since there's only one test utilizing `TESTS`, I'd just inline this into the parametrize block -- sorry I missed this on the first review

.pre-commit-hooks.yaml
```diff
@@ -105,6 +105,12 @@
     entry: end-of-file-fixer
     language: python
     files: \.(asciidoc|adoc|coffee|cpp|css|c|ejs|erb|groovy|h|haml|hh|hpp|hxx|html|in|j2|jade|json|js|less|markdown|md|ml|mli|pp|py|rb|rs|R|scala|scss|sh|slim|tex|tmpl|ts|txt|yaml|yml)$
+-   id: file-contents-sorter
+    name: File Contents Sorter
+    description: Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input in your .pre-commit-config.yaml file.
```
asottile: more specifically, they'd set the `files` key in the configuration

asottile:
Thanks!

dlgallagher:
Thank you for the quick back-and-forth. A good 1st github experience. Makes me want to try doing pull requests for some other tools and stuff we've used every day. 

Anyway, I'm going to go update hooks & set this one up in all our services.

asottile:
fwiw (unless things have changed) `deploy-blacklist.txt` is only necessary for classic soa.  Also, you may want to add a `^` to the beginning and a `$` to the end of your regex.

dlgallagher:
👍 on the regex change.
Your memory is spot-on; now that I look I don't see deploy-blacklist.txt used and the FE service generator doesn't have it so should be safe to delete. Thanks!

PR #203, opened 2017-06-23
Add sort-simple-yaml hook

dlgallagher:
Open source version of closed-source hook written by @chriskuehl 

> Sorts simple YAML files which consist only of top-level keys, preserving comments and blocks.


asottile:
Thanks! CC @chriskuehl

PR #204, opened 2017-06-23
Make hooks.yaml a symlink of .pre-commit-hooks.yaml

dlgallagher:
I was getting annoyed having to `cp .pre-commit-hooks.yaml hooks.yaml` each time I added a new hook. In yelp closed-source hook repo hooks.yaml is a symlink to .pre-commit-hooks.yaml so I figured it's probably OK here. 

`ln -s .pre-commit-hooks.yaml hooks.yaml`

In this pull request the diff makes it look like the file is deleted; perhaps github doesn't have a nice way to show a type change?


chriskuehl:
The reason we don't do this is because it doesn't work on windows :(

dlgallagher:
@chriskuehl aha thanks for the sanity check on that.

asottile:
Here's the full reasoning if curious: https://github.com/pre-commit/pre-commit/pull/470#issuecomment-274305200

dlgallagher:
@asottile thanks for the link for background/context.

PR #206, opened 2017-06-25
Some style tweaks (requirements_txt_fixer)

dlgallagher:
@asottile pointed out that I'd used a list where I could've used a tuple in a different PR. I changed that, then:

1. Made constants PASS=0/FAIL=1 since I found that made the code/tests much more readable
2. Inline the parameterized test params
3. Change an .append() loop to use .extend()


asottile:
I'm not sure `PASS` / `FAIL` improve readability -- they don't make it worse though so -0 I guess :)

asottile:
Thanks for the PR!

PR #207, opened 2017-06-25
Fix bug with the file-contents-sorter hook when processing file that …

dlgallagher:
…does not end in a newline

Discovered that when the file-contents-sorter gets a file with contents like: `b\na` (note that there is no newline at the end) it would rearrange them to result in `ab\n` which just looks like 'ab' on one line and is no-good. 

This branch fixes that by:

1. Stripping the line endings off each line from the start
2. Do the sorting
3. Add \n to the end of each line (including the last one)

Rather than comparing the str-ified before vs after to determine if the hook made changes, it compares the list-of-stripped-lines before vs after so this hook won't add newlines to the end of files if they do not actually require sorting. 

Moar test cases. 

Also, 
```
after_string = b'\n'.join(after) + b'\n'
f.write(after_string)
```
could just be:
`f.writelines("%s\n" % l for l in after)`
Except this repo runs tests in py34. The above works in 2.7 and py3.5 added bytestr formatting like this back in. Anyway, just including this rationale in case you wonder why I didn't just use writelines().

### DIFF
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -18,19 +18,18 @@ FAIL = 1
 
 
 def sort_file_contents(f):
-    before = tuple(f)
+    before = [line.strip(b'\n\r') for line in f if line.strip()]
     after = sorted(before)
 
-    before_string = b''.join(before)
-    after_string = b''.join(after)
-
-    if before_string == after_string:
+    if before == after:
         return PASS
-    else:
-        f.seek(0)
-        f.write(after_string)
-        f.truncate()
-        return FAIL
+
+    after_string = b'\n'.join(after) + b'\n'
+
+    f.seek(0)
+    f.write(after_string)
+    f.truncate()
+    return FAIL
 
 
 def main(argv=None):
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -18,18 +18,19 @@ FAIL = 1
 
 
 def sort_file_contents(f):
-    before = [line.strip(b'\n\r') for line in f if line.strip()]
-    after = sorted(before)
+    before = list(f)
+    after = sorted([line.strip(b'\n\r') for line in before if line.strip()])
 
-    if before == after:
-        return PASS
-
+    before_string = b''.join(before)
     after_string = b'\n'.join(after) + b'\n'
 
-    f.seek(0)
-    f.write(after_string)
-    f.truncate()
-    return FAIL
+    if before_string == after_string:
+        return PASS
+    else:
+        f.seek(0)
+        f.write(after_string)
+        f.truncate()
+        return FAIL
 
 
 def main(argv=None):
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Add this to your `.pre-commit-config.yaml`
   with single quoted strings.
 - `end-of-file-fixer` - Makes sure files end in a newline and only a newline.
 - `fix-encoding-pragma` - Add `# -*- coding: utf-8 -*-` to the top of python files.
-- `file-contents-sorter` - Sort the lines in specified files (defaults to alphabetical). You must provide list of target files as input to it.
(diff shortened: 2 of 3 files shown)


pre_commit_hooks/file_contents_sorter.py
```diff
@@ -18,19 +18,18 @@
 
 
 def sort_file_contents(f):
-    before = tuple(f)
+    before = [line.strip(b'\n\r') for line in f if line.strip()]
```
asottile: wont this remove blank lines?

dlgallagher: Yes it will, good catch!! Hmm if it rewrites file it should remove blank lines (otherwise they'd just end up all sorted together which is useless) but it shouldn't rewrite the file just to remove blank lines. I'll add test case for this with the fix.

asottile: `requirements-txt-sorter` handles this by grouping comment / blank lines for each "meaty" line and preserving them through sorting.  straight `sort` (while simpler) will put all of the comments / blank lines together (at the top? (not at a terminal to check))

dlgallagher: Oh wait actually I don't think this is an issue after all. We compare before vs after to determine whether file needs to be updated. However, the `before` list has blank lines removed by this list comprehension, so we're actually comparing before-without-blank-lines to after-without-blank-lines.

So if the input file has the lines in order but with extra lines scattered throughout, this code will see that file as passing and not edit it. 

So I don't think any changes are required here... did I miss something or does ^ make sense?

asottile: As written it would rewrite this file:

```
# hello
world

# down
here
```

to this
```
# down
# hello
here
world
```

(instead of)

```
# down
here
# hello
world
```

which seems non-ideal (whereas `requirements-txt-sorter` would preserve the chunking and comments)

dlgallagher: Ahh I see; I was intending this to be a simple general file sorter not specifically for python and as such should not respect a particular language's commenting. i.e. a deploy-whitelist.txt file I don't think people would expect python comments to be treated differently.

asottile: fwiw `frontend-service-make` respects `#` comments in `deploy-whitelist.txt`. 

Though I guess this is fine -- it should be documented in the readme that this sorting hook **will** remove blank lines and **won't** respect comments (unlike `requirements-txt-sorter`)

dlgallagher: OK I'll add that info to the readme 👍

asottile: I also think (to be consistent) even if it does not need to perform any sorting it should still be removing blank lines and ensuring the file ends in a newline -- that way one could run this hook and then not expect *unexpected* changes to their files later

dlgallagher: OK sure - the use case I want this for it doesn't matter either way and having it remove blank lines/end file in newline 100% of the time makes the code simpler. Updated!

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -18,19 +18,19 @@
...
+    after_string = b'\n'.join(after) + b'\n'
 
     if before_string == after_string:
         return PASS
-    else:
-        f.seek(0)
-        f.write(after_string)
-        f.truncate()
-        return FAIL
+
```
asottile: I prefer the lhs here, despite the `else` being superfluous, it is *obvious* at a glance that this code has two disparate branches.

dlgallagher:
Ok I added the `else` back in. 
Then did `git rebase -i 5dd1819e8b6757490f8c20b51df219d714d4ab0e` and squashed the most recent 3 commits down into 1 (my first time using git rebase). I got concerned when it wouldn't let me push after that, but googling around I found SO posts saying to force push so I did that.

pre_commit_hooks/file_contents_sorter.py
dlgallagher: OK; Added the `else` back in.

asottile:
Thanks again!

dlgallagher:
\o/ thanks @asottile!

PR #213, opened 2017-07-05
Report full version in check-ast

### LINKED ISSUE #191: Report exact python version in case of check-ast parse error
The `check-ast` hook only reports the executable name in case of a parse error. That is however sometimes not very helpful since it's mostly `python2` and `python3`.

I propose to use `sys.version.partition(" ")[0]` instead of `interpreter` in https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/check_ast.py#L26

asottile:
Resolves #191

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -10,6 +10,8 @@ omit =
     get-git-lfs.py
 
 [report]
+show_missing = True
+skip_covered = True
 exclude_lines =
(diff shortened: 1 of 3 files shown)


asottile:
Thanks @ulope for the code used here :D

PR #217, opened 2017-07-17
Add test instructions

nagromc:


nagromc:
Just realized there is https://github.com/pre-commit/pre-commit/blob/master/CONTRIBUTING.md

FML

Maybe some parts can be merged? Especially the parts where the existing Git configuration interfere with the test runner. I've been quite annoyed by this problem.

asottile:
While I think this is all correct and useful information, a lot of it is duplicated with the general tox documentation and CONTRIBUTING in pre-commit/pre-commit.

Note also that if you find your git configuration conflicting with the tests, [we consider it a bug](https://github.com/pre-commit/pre-commit-hooks/pull/212), [that should be fixed](https://github.com/pre-commit/pre-commit-hooks/pull/121) and would welcome changes to improve the testsuite or hooks themselves to be more resistant to alternate git configurations :)

I'm also not sure enumerating the supported python versions is necessary or useful as that information is already in [setup.py](https://github.com/pre-commit/pre-commit-hooks/blob/19b7f0f6ed583f3dda7062e0d272b07e02dfb8be/setup.py#L16-L20) and [tox.ini](https://github.com/pre-commit/pre-commit-hooks/blob/19b7f0f6ed583f3dda7062e0d272b07e02dfb8be/tox.ini#L4) (two standard places to search for this information).

asottile:
We'll handle this in https://github.com/pre-commit/pre-commit/issues/564

PR #228, opened 2017-08-09
Add mypy pre-commit hook

sjaensch:
The idea here is to use it primarily to make sure Python code is annotated properly. An example minimal configuration could look like this:

    [mypy]
    check_untyped_defs = True
    disallow_untyped_calls = False
    disallow_untyped_defs = True
    follow_imports = silent
    ignore_missing_imports = True

We've been using this hook internally and it works well. I've replaced the internal hook by pointing to my pre-commit-hooks fork and it continued to work.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -191,6 +191,15 @@
     # for backward compatibility
     files: ''
     minimum_pre_commit_version: 0.15.0
+-   id: mypy
+    name: mypy
+    description: This hook runs mypy.
+    entry: mypy
+    language: python
+    types: [python]
+    # for backward compatibility
+    files: ''
+    minimum_pre_commit_version: 0.15.0
 -   id: name-tests-test
     name: Tests should end in _test.py
     description: This verifies that test files are named correctly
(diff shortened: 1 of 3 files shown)


asottile:
A few issues:

- This is listed as a `language: python` hook, which means it gets its own virtual environment but I don't see `mypy` being added to `setup.py` to install it
- I believe `mypy` needs access to all of the installed dependencies in order to do proper type inference, as such it's better suited for either:
    - A test (line in `tox.ini`) where all dependencies are available
    - A `local` hook which assumes the virtualenv is active (technically, this is what is happening with the definition in this PR by luck), similar to the example given for `pylint` [here](http://pre-commit.com/#repository-local-hooks)
- If it is *intentionally* depending on an activated virtualenv, the `language: system` type is more appropriate (and doesn't require a noop virtualenv)
- imo duplicating docs about `files` and `args` (and `exclude`) is unnecessary in the README, this is covered in the [pre-commit framework documentation itself](http://pre-commit.com)
- It may be better to contribute to / use [existing mypy support for pre-commit](https://github.com/jerome-baek/pre-commit-mypy)

sjaensch:
Yup this is definitely not about type checking the codebase - as you pointed out, doing this through tox when you run tests is a much better option.

This hook's intention is to make sure all added / changed files are fully type annotated. That's why I added those lines to the README, although I agree I should document that better. You'll want to point to a pre-commit-specific config for mypy that does just that, and I provided an example of such a minimal config. The idea is that this hook can help you incrementally convert a bigger codebase to use type annotations, and spread the load of doing the annotations. Whenever a file is touched people are forced to type annotate it. Once your codebase is fully annotated you can then add some of the options I mentioned above to the mypy config that's used when you run tests - i.e. make sure all code that is checked is fully annotated.

sjaensch:
I'm going to use the repository you pointed out instead.

PR #231, opened 2017-08-16
handle file reading exceptions

TheKevJames:
I was going to just open an issue, but this turned out to be a relatively straightforward fix. See issue below; this PR ensures the filename is attached to the given errors.

Original issue: "Uncaught file-read exceptions hide filename from error message"

When `debug-statements` or `double-quote-string-fixer` fail to unicode decode a file, not even the filename is shown in the error output. More generally, any Exception being raised during parsing would lead to this behavior.

```
Debug Statements (Python)................................................Failed
hookid: debug-statements

  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/bin/debug-statement-hook", line 11, in <module>
    load_entry_point('pre-commit-hooks==0.9.1', 'console_scripts', 'debug-statement-hook')()
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/lib/python3.6/site-packages/pre_commit_hooks/debug_statement_hook.py", line 69, in debug_statement_hook
    retv |= check_file_for_debug_statements(filename)
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/lib/python3.6/site-packages/pre_commit_hooks/debug_statement_hook.py", line 38, in check_file_for_debug_statements
    ast_obj = ast.parse(open(filename).read(), filename=filename)
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/bin/../lib/python3.6/codecs.py", line 321, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf3 in position 55: invalid continuation byte

Fix double quoted strings................................................Failed
hookid: double-quote-string-fixer

  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/bin/double-quote-string-fixer", line 11, in <module>
    load_entry_point('pre-commit-hooks==0.9.1', 'console_scripts', 'double-quote-string-fixer')()
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/lib/python3.6/site-packages/pre_commit_hooks/string_fixer.py", line 70, in main
    return_value = fix_strings(filename)
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/lib/python3.6/site-packages/pre_commit_hooks/string_fixer.py", line 35, in fix_strings
    contents = io.open(filename).read()
  File "/Users/kevin/.pre-commit/repow04yynsa/py_env-python3.6/bin/../lib/python3.6/codecs.py", line 321, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf3 in position 55: invalid continuation byte
```

### DIFF
--- a/pre_commit_hooks/debug_statement_hook.py
+++ b/pre_commit_hooks/debug_statement_hook.py
@@ -42,6 +42,12 @@ def check_file_for_debug_statements(filename):
         print('\t' + traceback.format_exc().replace('\n', '\n\t'))
         print()
         return 1
+    except Exception:
+        print('{} - Caught exception while reading file'.format(filename))
+        print()
+        print('\t' + traceback.format_exc().replace('\n', '\n\t'))
+        print()
+        return 1
     visitor = ImportStatementParser()
     visitor.visit(ast_obj)
     if visitor.debug_import_statements:
--- a/pre_commit_hooks/string_fixer.py
+++ b/pre_commit_hooks/string_fixer.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import argparse
 import io
 import tokenize
+import traceback
 
 
 double_quote_starts = tuple(s for s in tokenize.single_quoted if '"' in s)
@@ -32,7 +33,14 @@ def get_line_offsets_by_line_no(src):
 
 
 def fix_strings(filename):
-    contents = io.open(filename).read()
+    try:
+        contents = io.open(filename).read()
+    except Exception:
+        print('{} - Caught exception while reading file'.format(filename))
+        print()
+        print('\t' + traceback.format_exc().replace('\n', '\n\t'))
+        print()
+        return 1
     line_offsets = get_line_offsets_by_line_no(contents)
 
     # Basically a mutable string


pre_commit_hooks/debug_statement_hook.py
```diff
@@ -42,6 +42,12 @@ def check_file_for_debug_statements(filename):
         print('\t' + traceback.format_exc().replace('\n', '\n\t'))
         print()
         return 1
+    except Exception:
+        print('{} - Caught exception while reading file'.format(filename))
+        print()
+        print('\t' + traceback.format_exc().replace('\n', '\n\t'))
+        print()
+        return 1
```
asottile: I think you'll find that this is already fixed: https://github.com/pre-commit/pre-commit-hooks/pull/230

pre_commit_hooks/string_fixer.py
```diff
@@ -32,7 +33,14 @@ def get_line_offsets_by_line_no(src):
 
 
 def fix_strings(filename):
-    contents = io.open(filename).read()
+    try:
+        contents = io.open(filename).read()
+    except Exception:
```
asottile: I'd like to see if we can't just fix this hook -- there's others in this repository that have been converted to working with arbitrarily-encoded files by only handling files as binary.

Please write a test for such a fix when attempted :)  You may find the approach in #230 relevant.

This also shouldn't blanket catch `Exception` -- this has the potential to hide actual programming errors

asottile:
Thanks for the PR!  I've left a few comments on the changes :)

pre_commit_hooks/debug_statement_hook.py
TheKevJames: That's remarkable timing on that fix. Looking forward to release v0.9.2!

pre_commit_hooks/string_fixer.py
TheKevJames: I tried my hand at this but ran into some issues. I'll keep playing with it at some point, but I'll post what I've come across already here:

We definitely need to read the file as bytes:

    contents = io.open(filename, 'rb').read()

That breaks casting to a list, but fortunately there's a byte-equivalent for that

    splitcontents = bytearray(contents)

That also means we need to tokenize differently; fortunately, `tokenize.tokenize` handles bytes:

    tokens = reversed(list(tokenize.tokenize(io.BytesIO(contents).readline, )))

When setting the new characters in the loop, we need to modify a `bytearray` now instead of a `list`. That'll be _something along the lines of_:

    splitcontents[...] = int.from_bytes(new_text.encode(), byteorder='big')

Finally, we can set

    new_contents = b''.join(splitcontents)

The above runs into a problem where somewhere along the lines I f*cked up and started tracking characters instead of lines (or something like that? In progress, still needs work, if anyone has any interest in carrying on with this _be my guest_).

asottile: Here's that in patch form:

```diff
diff --git a/pre_commit_hooks/string_fixer.py b/pre_commit_hooks/string_fixer.py
index 3523e8a..59b832f 100644
--- a/pre_commit_hooks/string_fixer.py
+++ b/pre_commit_hooks/string_fixer.py
@@ -11,15 +11,15 @@ double_quote_starts = tuple(s for s in tokenize.single_quoted if '"' in s)
 
 
 def handle_match(token_text):
-    if '"""' in token_text or "'''" in token_text:
+    if b'"""' in token_text or b"'''" in token_text:
         return token_text
 
     for double_quote_start in double_quote_starts:
         if token_text.startswith(double_quote_start):
             meat = token_text[len(double_quote_start):-1]
-            if '"' in meat or "'" in meat:
+            if b'"' in meat or b"'" in meat:
                 break
-            return double_quote_start.replace('"', "'") + meat + "'"
+            return double_quote_start.replace(b'"', b"'") + meat + b"'"
     return token_text
 
 
@@ -32,15 +32,15 @@ def get_line_offsets_by_line_no(src):
 
 
 def fix_strings(filename):
-    contents = io.open(filename).read()
+    contents = open(filename, 'rb').read()
     line_offsets = get_line_offsets_by_line_no(contents)
 
     # Basically a mutable string
-    splitcontents = list(contents)
+    splitcontents = bytearray(contents)
 
     # Iterate in reverse so the offsets are always correct
     tokens = reversed(list(tokenize.generate_tokens(
-        io.StringIO(contents).readline,
+        io.BytesIO(contents).readline,
     )))
     for token_type, token_text, (srow, scol), (erow, ecol), _ in tokens:
         if token_type == tokenize.STRING:
@@ -50,9 +50,9 @@ def fix_strings(filename):
                 line_offsets[erow] + ecol
             ] = new_text
 
-    new_contents = ''.join(splitcontents)
+    new_contents = bytes(splitcontents)
     if contents != new_contents:
-        with io.open(filename, 'w') as write_handle:
+        with open(filename, 'wb') as write_handle:
             write_handle.write(new_contents)
         return 1
     else:
diff --git a/tests/string_fixer_test.py b/tests/string_fixer_test.py
index a65213b..0603adf 100644
--- a/tests/string_fixer_test.py
+++ b/tests/string_fixer_test.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
@@ -49,3 +50,11 @@ def test_rewrite(input_s, output, expected_retval, tmpdir):
     retval = main([path.strpath])
     assert path.read() == output
     assert retval == expected_retval
+
+
+def test_non_utf8_file(tmpdir):
+    f_py = tmpdir.join('f.py')
+    f_py.write_binary('# -*- coding: cp1252 -*-\nx = "€"\n'.encode('cp1252'))
+    assert main((f_py.strpath,)) == 1
+    after = f_py.read_binary()
+    assert after == "# -*- coding: cp1252 -*-\nx = '€'\n".encode('cp1252')
```

This passes the tests in python 2, but I don't think `tokenize` works for bytes in python 3... hmm...  Instead the code could do the encoding dance (parse the comment cookie, assume utf-8 otherwise, decode to text, encode to the detected encoding) maybe?

asottile: Hmmm, actually the stdlib *does* support byte tokenizing:

## python2
- `tokenize.generate_tokens` supports both bytes and text

## python3
- `tokenize.tokenize` supports only bytes, the first token will be an `ENCODING` token which should be used to re-encode the result
- each token will have `str` (text) contents

So this is workable without using bytes throughout I think

I wonder if I could improve [tokenize-rt](https://github.com/asottile/tokenize-rt) to handle bytes to normalize these differences (and then use it here)

pre_commit_hooks/debug_statement_hook.py
asottile: This part is now available in 0.9.2 \o/

PR #241, opened 2017-10-07
Add hook for ensuring vcs permalinks

asottile:


pre_commit_hooks/check_vcs_permalinks.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import unicode_literals
+
+import argparse
+import re
+import sys
+
+
+GITHUB_NON_PERMALINK = re.compile(
+    b'https://github.com/[^/]+/[^/]+/blob/master/[^# ]+#L\d+',
+)
```
chriskuehl: is it also worth checking for something like `https://github.com/pre-commit/pre-commit-hooks/tree/master/pre_commit_hooks` vs `https://github.com/pre-commit/pre-commit-hooks/tree/dec98f7bc860eb34889eab98bcc8eebd2a379b62/pre_commit_hooks`?

asottile: I actually intentionally allow links to paths and files as those are much less likely to rot and often you want to link to the latest version.

PR #249, opened 2017-11-26
Add check to enforce literal syntax for Python builtin types

benwebber:
This check requires authors to initialize empty or zero builtin types
using the literal syntax (e.g., `{}` instead of `dict()`).

Authors may ignore this requirement for certain builtins using the
`--ignore` option.

Authors may also forbid calling `dict()` with keyword arguments
(`dict(a=1, b=2)`) using the `--no-allow-dict-kwargs` flag.


asottile:
Looks great!  Thanks for the PR :)

PR #251, opened 2017-11-30
check-builtin-literals: Ignore function attribute calls

benwebber:
Fixes an oversight in #249.

The AST differentiates between bare function calls (names) and attribute calls.


testing/resources/builtin_constructors.py
```diff
@@ -1,7 +1,18 @@
+# flake8 checks imports; `builtins` does not exist in Python 2.
```
asottile: `from six.moves import builtins` will appease flake8 -- let's do that instead :)

asottile:
Oops forgot to merge this. Thanks again!

PR #254, opened 2017-12-10
Refactor/pretty format json

cas--:
Two commits to fix pretty_format_json:

**Remove pretty_format_json simplejson dependency**

 * The simplejson module is only needed for <=py25 so replace with builtin json.
 * Replace six dependecy for simple Py2 check for convertion to unicode.
 * Cleanup quotes.

----
**Fix pretty_format_json to use int indent**

The indent parameter for json should be integer and under Python2 is
will raise an error if not. So switch from str to int and mention
default value in help text.

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -33,24 +33,12 @@ def _autofix(filename, new_contents):
         f.write(new_contents)
 
 
-def parse_indent(s):
-    # type: (str) -> str
+def parse_num_to_int(s):
+    """Convert string numbers to int, leaving strings as is."""
     try:
-        int_indentation_spec = int(s)
+        return int(s)
     except ValueError:
-        if not s.strip():
-            return s
-        else:
-            raise ValueError(
-                'Non-whitespace JSON indentation delimiter supplied. ',
-            )
-    else:
-        if int_indentation_spec >= 0:
-            return int_indentation_spec * ' '
-        else:
-            raise ValueError(
-                'Negative integer supplied to construct JSON indentation delimiter. ',
-            )
+        return s
 
 
 def parse_topkeys(s):
@@ -68,9 +56,12 @@ def pretty_format_json(argv=None):
     )
     parser.add_argument(
         '--indent',
-        type=parse_indent,
-        default='  ',
-        help='String used as delimiter for one indentation level',
+        type=parse_num_to_int,
+        default='2',
+        help=(
+            'The number of indent spaces or a string to be used as delimiter'
+            ' for indentation level e.g. 4 or "\t" (Default: 2)'
+        ),
     )
     parser.add_argument(
         '--no-ensure-ascii',
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -2,11 +2,11 @@ from __future__ import print_function
 
 import argparse
 import io
+import json
 import sys
 from collections import OrderedDict
 
-import simplejson
-import six
+from six import text_type
 
 
 def _get_pretty_format(contents, indent, ensure_ascii=True, sort_keys=True, top_keys=[]):
@@ -17,18 +17,18 @@ def _get_pretty_format(contents, indent, ensure_ascii=True, sort_keys=True, top_
         if sort_keys:
             after = sorted(after, key=lambda x: x[0])
         return OrderedDict(before + after)
-    return six.text_type(simplejson.dumps(
-        simplejson.loads(
-            contents,
-            object_pairs_hook=pairs_first,
-        ),
+    json_pretty = json.dumps(
+        json.loads(contents, object_pairs_hook=pairs_first),
         indent=indent,
         ensure_ascii=ensure_ascii,
-    )) + "\n"  # dumps does not end with a newline
+        separators=(',', ': '),  # Workaround for https://bugs.python.org/issue16333
+    )
+    # Ensure unicode (Py2) and add the newline that dumps does not end with.
+    return text_type(json_pretty) + '\n'
 
 
 def _autofix(filename, new_contents):
-    print("Fixing file {}".format(filename))
+    print('Fixing file {}'.format(filename))
     with io.open(filename, 'w', encoding='UTF-8') as f:
         f.write(new_contents)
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -28,8 +28,6 @@
         'flake8!=2.5.3',
         'autopep8>=1.3',
         'pyyaml',
-        'simplejson',
-        'six',
```
asottile: six is used elsewhere so it can't be removed

pre_commit_hooks/pretty_format_json.py
```diff
@@ -110,16 +110,16 @@ def pretty_format_json(argv=None):
...
-                print("File {} is not pretty-formatted".format(json_file))
+                print('File {} is not pretty-formatted'.format(json_file))
 
                 if args.autofix:
                     _autofix(json_file, pretty_contents)
 
                 status = 1
 
-        except simplejson.JSONDecodeError:
+        except json.JSONDecodeError:
```
asottile: there's no `json.JSONDecodeError`, I believe stdlib throws `ValueError`?

pre_commit_hooks/pretty_format_json.py
```diff
@@ -17,36 +15,38 @@ def pairs_first(pairs):
...
+    json_pretty = json.dumps(
+        json.loads(contents, object_pairs_hook=pairs_first),
         indent=indent,
         ensure_ascii=ensure_ascii,
-    )) + "\n"  # dumps does not end with a newline
+    )
+    # Ensure unicode under Py2
+    if sys.version_info.major == 2:
+        json_pretty = json_pretty.decode()
+    json_pretty += '\n'  # dumps do not end with a newline
```
asottile: the comment was changed here, though I think the lhs has better grammar

pre_commit_hooks/pretty_format_json.py
```diff
@@ -17,36 +15,38 @@ def pairs_first(pairs):
...
-            object_pairs_hook=pairs_first,
-        ),
+    json_pretty = json.dumps(
+        json.loads(contents, object_pairs_hook=pairs_first),
         indent=indent,
         ensure_ascii=ensure_ascii,
-    )) + "\n"  # dumps does not end with a newline
+    )
+    # Ensure unicode under Py2
+    if sys.version_info.major == 2:
```
asottile: will need a `no cover` comment to satisfy coverage.

Since six is still used, probably just use `six.PY2` here

pre_commit_hooks/pretty_format_json.py
```diff
@@ -69,8 +69,8 @@ def pretty_format_json(argv=None):
     parser.add_argument(
         '--indent',
         type=parse_indent,
-        default='  ',
-        help='String used as delimiter for one indentation level',
+        default=2,
+        help='String used as delimiter for one indentation level (Default: 2)',
```
asottile: this shouldn't be refactored -- it allows for the following inputs:

- `'\t'`
- '  '
- '4'

See https://github.com/pre-commit/pre-commit-hooks/pull/120 which is essentially the opposite of this patch :)

pre_commit_hooks/pretty_format_json.py
cas--: Yeah intentional :)

pre_commit_hooks/pretty_format_json.py
cas--: Hmm ok, strings break Py2 json. Suggestions? Keep simplejson and have json as fallback?

There is no need for that function tbh since both simplejson and py3 json accept both str and int and do the conversion...

setup.py
cas--: I missed that other usage, that can be easily resolved too. :)

pre_commit_hooks/pretty_format_json.py
cas--: The original code never allowed `'\t'`. `'\t'.strip() == '\t'`

`''` and `'4'` are still value and coverted to 0 and 4 ints.

cas--: Yes and `\t` raised a valueerror with original code.

cas--: Ok what I suggest is to leave the default as an int and just pass the values `str` or `int` to json and catch any exceptions.

pre_commit_hooks/pretty_format_json.py
cas--: Ah it does exist on Py3 but not Py2... I'll sort this :+1:

cas--:
btw I thought I had run the tests before creating the PR, sorry about that!

pre_commit_hooks/pretty_format_json.py
asottile:
```python
>>> '\t'.strip()
''
```

hmmm seems to do what I expect :)

asottile: Oh I see you edited, disregard :D

pre_commit_hooks/pretty_format_json.py
asottile: I guess it makes more sense to me that `dumps` is a singular noun ("dump string")

pre_commit_hooks/pretty_format_json.py
cas--: Yeah I was testing with strip(' ') not strip() :scream: 

I think I have the solution for all this and will update...

setup.py
asottile: tbh, until we're comfortable dropping python 2 entirely, keeping six around seems ok to me -- I'd certainly rather see `from six.moves import ...` than `try: except:` for imports

cas--: For a single import to require six, I'd disagree as it creates an extra level of abstraction to a 3rd part module and a docs lookup for usage. With a try..except you know exactly what is going on. Both will have to be modified in future regardless though I suppose it'd then be easier to find the six module imports...

cas--:
This was a fun project... much head scratching with failing tests, found a bug in object_pairs_hook for py2.7 and 3.4: https://bugs.python.org/issue16333

Anyway all sorted, I think...

asottile:
ah yeah, I've hit that json bug before, fun fun!

asottile:
Thanks again!

PR #256, opened 2018-01-14
Update detect_private_key.py

nicain:
Add to blacklist of key matches, to reject key files generated by google cloud.  Addresses https://github.com/pre-commit/pre-commit-hooks/issues/255#issuecomment-357480631,
tested on this use-case and works.


asottile:
Thanks!

PR #257, opened 2018-01-15
Give pre-commit-python-sorter a new home

benjaoming:
Suggesting to move this because original team has stopped producing releases and people are running off third-party repos with funky sha's.

Thanks @Dinoshauer @teeberg @amureki @jeffshek @Tenzer @asottile 

Previous repo: https://github.com/FalconSocial/pre-commit-python-sorter/blob/master/setup.py

I know about https://github.com/asottile/reorder_python_imports but I think it's too complex for the official repo. This one is pretty short, and could easily have a nice+stable retirement home here :)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -247,6 +247,13 @@
     entry: requirements-txt-fixer
     language: python
     files: requirements.*\.txt$
+-   id: python-import-sorter
+    name: Sort python imports
+    description: This hook sorts python imports.
+    entry: python-import-sorter
+    language: python
+    files: \.py$
+    minimum_pre_commit_version: 0.15.0
 -   id: sort-simple-yaml
     name: Sort simple YAML files
     language: python
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Add this to your `.pre-commit-config.yaml`
 - `no-commit-to-branch` - Protect specific branches from direct checkins.
     - Use `args: -b <branch> ` to set the branch. `master` is the default if no argument is set.
 - `pyflakes` - Run pyflakes on your python files.
+- `python-import-sorter` - Sorts Python imports with isort.
 - `pretty-format-json` - Checks that all your JSON files are pretty.  "Pretty"
   here means that keys are sorted and indented.  You can configure this with
   the following commandline options:
--- a/hooks.yaml
+++ b/hooks.yaml
@@ -172,6 +172,12 @@
     entry: upgrade-your-pre-commit-version
     files: ''
     minimum_pre_commit_version: 0.15.0
+-   id: python-import-sorter
+    language: system
+    name: upgrade-your-pre-commit-version
+    entry: upgrade-your-pre-commit-version
+    files: ''
+    minimum_pre_commit_version: 0.15.0
 -   id: sort-simple-yaml
     language: system
     name: upgrade-your-pre-commit-version
--- a/pre_commit_hooks/sort_python_imports.py
+++ b/pre_commit_hooks/sort_python_imports.py
@@ -1,0 +1,69 @@
+"""
+Migrated from https://github.com/FalconSocial/pre-commit-python-sorter/
+by @benjaoming
+
+Thanks to FalconSocial devs for starting this.
+Original author: Kasper Jacobsen, @Dinoshauer
+
+The MIT License (MIT)
+
+Copyright (c) 2014 Falcon Social
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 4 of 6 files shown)


asottile:
First off, thanks for the interest in pre-commit and the PR :)

I think (correct me if I'm incorrect!) that the development of `pre-commit-python-sorter` mostly lost interest due to the ability to run `isort` directly as a [`local` hook](https://pre-commit.com/#repository-local-hooks) and that modifications to files automatically cancelled the commit https://github.com/pre-commit/pre-commit/pull/287.

That said, I do think there's interest in having a "pre-commit" managed usage of `isort` (at least until we can convince upstream to add the necessary metadata).  As such, I've gone ahead and set up [an isort mirror](https://github.com/pre-commit/mirrors-isort) which should keep up to date :)

As for this PR, if we separately want to maintain `pre-commit-python-sorter` (note that I also maintain `reorder_python_imports` ;)) I think we should do so in a separate repository.  That is if we want to do this at all (I'm unclear on what value it adds over vanilla `isort` -- perhaps you or one of the original devs could elaborate on that!).

benjaoming:
> and that modifications to files automatically cancelled the commit pre-commit/pre-commit#287

That's pretty cool -- but I thought it was a feature TBH :)

benjaoming:
> I think we should do so in a separate repository.

My proposal here is that the feature is so popular, and the codebase is so simple, that it makes sense to centralize it for community to be able to contribute in a place that's well-maintained.

Can you add some subjective reflections on `reorder_python_imports`? Like do you think that `pre-commit-python-sorter` is useful in some cases? Or should it really just be replaced?

benjaoming:
...just to make it clear: I have no intention or bias to backup this, just wanted to open it to see if there was a pragmatic interest :) If `reorder_python_imports` should be the choice for everyone and the repo is healthy (and external), there's no need for this! I can also open up a PR on FalconSocial to suggest a link to `reorder_python_imports`.

asottile:
I guess I should clarify some long-term goals for `pre-commit/pre-commit-hooks` itself first :)

In the further out long term, `pre-commit/pre-commit-hooks` should just contain language-agnostic hooks.  The initial (and current) implementation contains some python-specific hooks which I imagine will eventually find a home at something like (does not exist yet) `pre-commit/pre-commit-hooks-python`

Some effort has already been made in that direction: [mirrors-autopep8](https://github.com/pre-commit/mirrors-autopep8) has been separated obviating the need for `autopep8-wrapper` https://github.com/pre-commit/pre-commit-hooks/issues/92 -- autopep8-wrapper was originally written because https://github.com/pre-commit/pre-commit/issues/285 was not implemented -- that is a hook could silently modify files (which should be staged for a commit) and would succeed.  Once pre-commit adjusted to make this a failure case, the usefulness of the wrapper diminished greatly.

In a lot of ways, `pre-commit/pre-commit-hooks` isn't far off from needing a split.  It has a bit of a sprawling-dependency problem and I don't really want to make that worse by adding yet-another dependency (when it isn't necessarily going to be used _most_ of the time).  Logically, I think `pre-commit/pre-commit-hooks` could have the following hooks split out to a python-specific repository _today_ and start the migration: `autopep8-wrapper` (though, probably deprecate), `check-ast`, `check-builtin-literals`, `check-docstring-first`, `debug-statements`, `double-quote-string-fixer`, `fix-encoding-pragma`, `flake8` + `pyflakes` (probably move to a mirror, or attempt to upstream the configuration), `name-tests-test`, `requirements-txt-fixer`.

Now onto `reorder_python_imports` vs `pre-commit-python-sorter`

I definitely agree that import sorting is by far one of the most important hooks for python repositories.  A lot of the motivation for using a framework like this is to get high quality fixers and use them!

`reorder_python_imports` (at the moment) largely supports a *very specific* style of imports -- the style that _I_ prefer and isn't very configurable (whether you consider this a plus or minus is up to interpretation 😆).  There is one core benefit that it currently has over `isort` though I think these could be fixed on `isort`'s side -- though I've tried in the past and didn't get far: https://github.com/timothycrosley/isort/issues/184.  This is that `isort` is pretty poor at classifying imports -- mostly relying on manual configuration and [a crazy long hardcoded list](https://github.com/timothycrosley/isort/blob/3b011412742d09ba768346379a52cd0784d1514f/isort/settings.py#L57-L103).  Here, `reorder_python_imports` (well, more accurately [`aspy.refactor_imports`](https://github.com/asottile/aspy.refactor_imports) the actual classifying implementation) excels by intelligently classifying imports based on project structure.

The other main difference (and in my mind a _minor_ difference) is `reorder_python_imports` is _much_ more timid about reordering imports when there is interspersed code.  I chose this design as I wanted import reordering to be as failproof as possible -- now there _are_ situations where imports *must* be ordered in a specific way and they'd still break under both implementations -- just `reorder_python_imports` is significantly more conservative.

On to `pre-commit-python-sorter`!  As far as I can tell from reading the source, this is just a simple wrapper around `isort`'s library functionality.  I _think_ (and may be incorrect in this assertion!) that it was written before (1) `isort` could be easily used directly (2) pre-commit would consider modifications a failure.   This necessitated a trivial wrapper which added this functionality -- and `pre-commit-python-sorter` was born!  As far as I can tell, `mirrors-isort` (or metadata added to `isort` itself) should completely obsolete this.

benjaoming:
Just my initial reaction to your response @asottile : THANKS!!! You've made a wonderful clarification, way more that I could ask for! I'm happily closing this PR and would at any time feel good about making another one. Thanks for being open and thanks for offering a wonderful project to the community! :heart:

asottile:
👍 And thank you for the PR!  I created `mirrors-isort` as a direct result of this issue, so improvements all around 🎉

Dinoshauer:
@asottile You're completely correct in both of your assumptions at the end of your post. God speed :)

Issue #258, opened 2018-01-20
detect_aws_credentials hook raises a false positive match when AWS credential values contain only a space character(s)


PVSec:
The detect_aws_credentials hook raises a false positive match when AWS credential values contain only a space character(s). For example, if the ~/.aws/credentials file contains the following:

```
[test]
aws_access_key_id = ggfgafg
aws_secret_access_key = 
```
By having a space character after the "=" for the _aws_secret_access_key_ variable it results in:
```
Detect AWS Credentials...................................................Failed
hookid: detect-aws-credentials

AWS secret found in test.txt: ****************************
```

Which is technically true, but in essence the hook should trim (remove) the space(s) in the values for the variables prior to checking the file(s) to prevent this confusing issue from occurring.

asottile:
👍 

would you like to submit a PR for this?

I think the fix just involves checking for an empty string while retrieving from https://github.com/pre-commit/pre-commit-hooks/blob/0ddb3b8527b29e56ce97e5e8308b7a871f53b3a7/pre_commit_hooks/detect_aws_credentials.py#L56

asottile:
via #260

asottile:
This has been released as part of [v1.2.1](https://github.com/pre-commit/pre-commit-hooks/releases/tag/v1.2.1)!

(sorry about the slow turnaround, I thought I had released this already! 😆)

PR #260, opened 2018-01-26
Fix for issue #258

PVSec:
Adding a check to make sure either spaces or null values are not mistakenly added into the key variable


asottile:
Looks great!  Mind adding a quick test [here](https://github.com/pre-commit/pre-commit-hooks/blob/e19917ae62c365d4d746fa9639c7506435bcf1c8/tests/detect_aws_credentials_test.py#L64-L93)?

asottile:
Thanks for the PR!

PR #264, opened 2018-02-16
Add print-message pre-commit hook

fragosti:
We would like to surfaces messages for when certain files (matching a regex) are modified. This PR adds this pre-commit hook and additionally allows you to fail automatically when certain files are modified.

Do I need to add entries to hooks.yaml and the other configs as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ Add this to your `.pre-commit-config.yaml`
     - `--indent ...` - Control the indentation (either a number for a number of spaces or a string of whitespace).  Defaults to 4 spaces.
     - `--no-sort-keys` - when autofixing, retain the original key ordering (instead of sorting the keys)
     - `--top-keys comma,separated,keys` - Keys to keep at the top of mappings.
+- `print-message` - Print a message and optionally fail when triggered.
+    - `--message ...` - The message to print.
+    - `--fail` - Whether to fail the pre-commit check.
 - `requirements-txt-fixer` - Sorts entries in requirements.txt
 - `sort-simple-yaml` - Sorts simple YAML files which consist only of top-level keys, preserving comments and blocks.
 - `trailing-whitespace` - Trims trailing whitespace.
--- a/pre_commit_hooks/print_message.py
+++ b/pre_commit_hooks/print_message.py
@@ -1,0 +1,33 @@
+from __future__ import print_function
+
+import argparse
+import sys
+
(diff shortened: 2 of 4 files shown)


asottile:
Could you clarify the usecase a bit more?

Sounds to me like you actually want to use `.gitignore` to prevent writing to specific files.

You'll also want to play around with [`pre-commit try-repo`](https://pre-commit.com/#pre-commit-try-repo) to demonstrate how this'll work.

I also don't think `--fail` should be an option, otherwise the messages will just evaporate into the aether.

How will someone use this hook?  Seems to me like they'd have to _always_ override `files` / `types` *and* `args` for it to be useful.  The current hooks currently housed here all work out of the box without configuration.

asottile:
Another idea: can this be more easily written as a [repository `local` hook](https://pre-commit.com/#repository-local-hooks) and maybe leverage [`pygrep`](https://pre-commit.com/#pygrep)?  I _think_ those should be sufficient to satisfy your usecase without needing this hook.

fragosti:
I see. Yes the idea is that you'd always supply `files` and `args`. I can improve that by adding to the README?

I can give you a specific use-case. We actually don't need it to fail so I'm fine with removing that option. Basically it's just for developer awareness. We want to be able to target certain files to say they're prone to a certain type of bug and that they should consider manually testing in a certain way, or running a certain test suite in environments where not all test suites are run on a per-push basis (like our iphone project).

Update: looking into local hooks.

asottile:
Sounds to me like you're adding warning noise which doesn't really fit well with this repository (which is all linting / fixing).

My general stance is to discourage warning noise -- either it should fail outright, fix it, or do nothing.  By adding unnecessary (and sometimes constant!) output you'll train your users to ignore useful information.

pre-commit won't display hook output unless running with `--verbose` or if the hook fails anyway so I don't think this warning idea really works at all?

fragosti:
Ok fair enough! Trying to get this working on more of the review-branch level (in "Testing Done" for example) but was interested in using something like this in the meantime. I'm fine with closing the PR.

asottile:
Oh I see, you're at yelp!

Maybe you'll want to try something _like_ this in the closed source `yelp_pre_commit_hooks` repository first (if it still exists? I assume it still exists?) since I'm not yet convinced it's a good idea.  Should give you a good test bed and you can see whether the usecase works or not :)

fragosti:
Oh! Did not know that was a thing. Thanks!

asottile:
As above, closing this for now :)

PR #267, opened 2018-02-23
Validate code to check if datetime.now() is being used

adlermedrado:
Most of projects that I work on are using UTC datetime, so datetime.now() may cause problems.
This hook can help to avoid such cases.

### DIFF
--- a/pre_commit_hooks/detect_datetime_now.py
+++ b/pre_commit_hooks/detect_datetime_now.py
@@ -1,0 +1,24 @@
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import os
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(description='Check if datetime.now is being used.')
+    parser.add_argument('filenames', nargs='*', help='Filenames to check')
+    args = parser.parse_args(argv)
+
+    for filename in args.filenames:
+        if os.path.isfile(filename):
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/detect_datetime_now.py
```diff
@@ -0,0 +1,24 @@
...
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(description='Check if datetime.now is being used.')
+    parser.add_argument('filenames', nargs='*', help='Filenames to check')
+    args = parser.parse_args(argv)
+
+    for filename in args.filenames:
+        if os.path.isfile(filename):
+            with open(filename, 'rb') as f:
+                if b'datetime.now()' in f.read():
```
asottile: This can be done even simpler as a [`pygrep`](https://pre-commit.com/#pygrep) hook

This would also give better output (currently the hook fails but doesn't mention the filename or line).

asottile:
A sketch of what that might look like:


```yaml
-   id: check-datetime-now
    name: Prefer datetime.utcnow()
    language: pygrep
    entry: 'datetime\.now\(\)'
    types: [python]
```

pre_commit_hooks/detect_datetime_now.py
adlermedrado: Thank you. Your suggestion looks simpler.

asottile:
@adlermedrado cheers!  Thanks for the PR :)

PR #275, opened 2018-03-25
Provide automatic removal of pkg-resources==0.0.0

nvtkaszpir:
Should help to deal with that pretty paintuly issue under Ubuntu/Debian
family:
https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1635463

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -72,8 +72,8 @@ def fix_requirements(f):
     # find and remove pkg-resources==0.0.0
     # which is automatically added by broken pip package under Debian
     requirements = [
-        requirement for requirement in requirements
-        if requirement.value != b'pkg-resources==0.0.0\n'
+        req for req in requirements
+        if req.value != b'pkg-resources==0.0.0\n'
     ]
 
     for requirement in sorted(requirements):
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -71,9 +71,10 @@ def fix_requirements(f):
 
     # find and remove pkg-resources==0.0.0
     # which is automatically added by broken pip package under Debian
-    for requirement in requirements:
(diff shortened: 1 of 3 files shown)


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -69,6 +69,11 @@ def fix_requirements(f):
     else:
         rest = []
 
+    for requirement in requirements:
+        if b'pkg-resources' in requirement.name:
+            if b'0.0.0' in requirement.value:
```
asottile: I think this makes more sense?

```python
if requirement.value == b'pkg-resources==0.0.0\n':
    ...
```

asottile:
Otherwise looks good, shame debian botched the metadata here :(

pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -69,10 +69,11 @@ def fix_requirements(f):
...
         rest = []
 
+    # find and remove pkg-resources==0.0.0
+    # which is automatically added by broken pip package under Debian
     for requirement in requirements:
-        if b'pkg-resources' in requirement.name:
-            if b'0.0.0' in requirement.value:
-                requirements.remove(requirement)
+        if requirement.value == b'pkg-resources==0.0.0\n':
+            requirements.remove(requirement)
```
asottile: actually, now that I look at it I think this loop results in UB (modifying list while iterating over it) -- you can avoid this with:

```python
requirements = [
    requirement for requirement in requirements
    if requirement.value != b'pkg-resources==0.0.0\n'
]
```

asottile:
Thanks for the feature!

nvtkaszpir:
Thanks for the comments. Looking at it now all those comments and CI failures ... it's a sign I should go to sleep ;)

asottile:
This has been (finally!) released as part of [v1.3.0](https://github.com/pre-commit/pre-commit-hooks/releases/tag/v1.3.0)! 🎉

PR #276, opened 2018-03-25
Ensure tox gets proper language settings.

nvtkaszpir:
If LC_ALL is not passed then the interpreter will assume it is C
(which causes the interpreter to use ASCII). This, in turn, causes
the unicode in various places to trigger a UnicodeDecodeError
exception.

CI workers have that locale set by default.
Some people do not have that locale setting set properly.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@ setenv =
     GIT_COMMITTER_NAME = "test"
     GIT_AUTHOR_EMAIL = "test@example.com"
     GIT_COMMITTER_EMAIL = "test@example.com"
+	LC_ALL=en_US.UTF-8
 commands =
     coverage erase
     coverage run -m pytest {posargs:tests}


asottile:
A couple things:
- you've got tabs ;)
- I actually think that assuming a `UTF-8` locale is probably a bug, if there's things that are failing they should probably get fixed?  Do you have some specific stacktraces here -- I imagine they'd make decent bug reports :)
- I think `LANG` is probably a better catchall than `LC_ALL` (iirc, `LANG` overrides all of the others)

nvtkaszpir:
- argh...
- example error
```
_______________________________________________________________________________________________________ test_readme_contains_all_hooks ________________________________________________________________________________________________________

    def test_readme_contains_all_hooks():
>       readme_contents = io.open('README.md').read()

tests/readme_test.py:10: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <encodings.ascii.IncrementalDecoder object at 0x00000000050dcaa0>
input = "[![Build Status](https://travis-ci.org/pre-commit/pre-commit-hooks.svg?branch=master)](https://travis-ci.org/pre-comm...d like to use these hooks, they're also available as a standalone\npackage.\n\nSimply `pip install pre-commit-hooks`\n"
final = 1

    def decode(self, input, final=False):
>       return codecs.ascii_decode(input, self.errors)[0]
E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1828: ordinal not in range(128)

.tox/pypy/lib-python/2.7/encodings/ascii.py:26: UnicodeDecodeError

```
- they work in the same way, but LC_ALL is processed in the first place, see ``man 7 locale``. If you suggest to stick to LANG then OK.

asottile:
Ah indeed! Looks to me like a buggy test, I believe this fixes it? https://github.com/pre-commit/pre-commit-hooks/pull/277

nvtkaszpir:
Yep.
So my PR is no longer required. :)

asottile:
via #279 -- thanks again :D

PR #287, opened 2018-05-18
Detect putty & sshcom private keys

vin01:
Unfortunately private key format is not standardized like [public key format for SSH](https://tools.ietf.org/html/rfc4716) allowing different tools/vendors to use different formats. 

**Motivation:** I noticed it with putty keys while using `detect_private_key` hook and a little research showed that there does not seem to be a standard so I have added header lines for putty & ssh.com private keys.


asottile:
Thanks!

asottile:
This has been (finally!) released as part of [v1.3.0](https://github.com/pre-commit/pre-commit-hooks/releases/tag/v1.3.0)! 🎉

PR #292, opened 2018-06-04
Update regex for VCS permalinks

mxr:
Support insecure links and links with `www`

### DIFF
--- a/pre_commit_hooks/check_vcs_permalinks.py
+++ b/pre_commit_hooks/check_vcs_permalinks.py
@@ -8,7 +8,7 @@ import sys
 
 
 GITHUB_NON_PERMALINK = re.compile(
-    b'https://github.com/[^/]+/[^/]+/blob/master/[^# ]+#L\d+',
+    b'https?://(www\.)?github.com/[^/]+/[^/]+/blob/master/[^# ]+#',
 )
 
 
--- a/tests/check_vcs_permalinks_test.py
+++ b/tests/check_vcs_permalinks_test.py
@@ -1,6 +1,8 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import pytest
+
 from pre_commit_hooks.check_vcs_permalinks import main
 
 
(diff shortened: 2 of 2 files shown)


pre_commit_hooks/check_vcs_permalinks.py
```diff
@@ -8,7 +8,7 @@
 
 
 GITHUB_NON_PERMALINK = re.compile(
-    b'https://github.com/[^/]+/[^/]+/blob/master/[^# ]+#L\d+',
+    b'https?://(www\.)?github.com/[^/]+/[^/]+/blob/master/[^# ]+#',
```
asottile: github redirects `http` and `www.github.com` to `https://github.com`, so I don't think this is worth covering / complicating

tests/check_vcs_permalinks_test.py
```diff
@@ -20,17 +22,27 @@ def test_passing(tmpdir):
...
 
-def test_failing(tmpdir, capsys):
+@pytest.mark.parametrize(
+    'contents',
+    (
+        'http://github.com/asottile/test/blob/master/foo#L1',
+        'http://www.github.com/asottile/test/blob/master/foo#L1',
+        'https://github.com/asottile/test/blob/master/foo#L1',
+        'https://www.github.com/asottile/test/blob/master/foo#L1',
+        'https://www.github.com/asottile/test/blob/master/foo#my-anchor',
```
asottile: This arguably _shouldn't_ be a permalink.  For example https://github.com/asottile/pyupgrade/blob/master/README.md#dictionary-comprehensions

mxr: Can you clarify what you mean? https://github.com/asottile/pyupgrade/blob/a8bdf45b6b91b8465cc739d350d56c7fb260a8db/README.md#dictionary-comprehensions is a valid link

mxr: (I hit submit on the comment too fast) I was thinking that someone could remove the anchor in a future commit

pre_commit_hooks/check_vcs_permalinks.py
mxr: Yes, but if you check in `http://github.com/asottile/test/blob/master/foo#L1` into a codebase, then you would check in a link that could break in the future. The regex before my change doesn't catch that case

tests/check_vcs_permalinks_test.py
asottile: The hook is intended to catch links to code, if documentation breaks headers that's probably a problem with the documentation (I would, for instance, link to [latest:parametrize](https://docs.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids) preferably over [2.7.3:parametrize](https://docs.pytest.org/en/2.7.3/example/parametrize.html#different-options-for-test-ids) (old docs))

The point here being that linking to a specific line of code against a master branch is _always_ wrong, but linking to a header on some documentation is most of the time just fine.

The general design of hooks I maintain aim to *minimize* false-positives

pre_commit_hooks/check_vcs_permalinks.py
asottile: where would you get such a link? manually typed out?

asottile: That said, this is by far the less controversial part of this patch -- split out the other part and I can accept this

tests/check_vcs_permalinks_test.py
mxr: Gotcha, thanks for the context

pre_commit_hooks/check_vcs_permalinks.py
mxr: Hmm good point. Will think about this. Thinking more, you could drop the `https?://(www\.)?` part entirely in the regex (it would transparently catch the same scenarios)

mxr:
Closing per the above discussion

pre_commit_hooks/check_vcs_permalinks.py
mxr: Hmm OTOH if someone has a line length limit they might drop the `https://` 🤔

PR #293, opened 2018-06-04
Improve vcs regex (don't match whitespace)

asottile:


mxr:
Microsoft allows spaces in usernames & repos?

asottile:
No they don't, that's why I made the regex better :D

PR #299, opened 2018-06-21
fix two words which got swapped.

bobbyrullo:


asottile:
@bobbyrullo also note there's a new~ish `--unsafe` argument which'll enable roughly the same thing

PR #301, opened 2018-06-26
Added --assume-in-merge option for check-merge-conflict

### LINKED ISSUE #300: Option to force run check-merge-conflict
check-merge-conflict only runs when it detects merge/rebase in progress.
This means it will always pass on CI setups.

Am not sure if this should be fixed specifically for check-merge-conflict or make a generic CI functionality available via [pre-commit](https://github.com/pre-commit/pre-commit/)

vinayinvicible:
Fixes #300

### DIFF
--- a/pre_commit_hooks/check_merge_conflict.py
+++ b/pre_commit_hooks/check_merge_conflict.py
@@ -26,9 +26,10 @@ def is_in_merge():
 def detect_merge_conflict(argv=None):
     parser = argparse.ArgumentParser()
     parser.add_argument('filenames', nargs='*')
+    parser.add_argument('--assume-in-merge', action='store_true')
     args = parser.parse_args(argv)
 
-    if not is_in_merge():
+    if not is_in_merge() and not args.assume_in_merge:
         return 0
 
     retcode = 0
--- a/tests/check_merge_conflict_test.py
+++ b/tests/check_merge_conflict_test.py
@@ -128,5 +128,12 @@ def test_ignores_binary_files():
 
 
 def test_does_not_care_when_not_in_a_merge(tmpdir):
-    tmpdir.join('README.md').write('problem\n=======\n')
-    assert detect_merge_conflict(['README.md']) == 0
+    f = tmpdir.join('README.md')
+    f.write_binary(b'problem\n=======\n')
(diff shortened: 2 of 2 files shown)


pre_commit_hooks/check_merge_conflict.py
```diff
@@ -12,23 +12,26 @@
...
 
 
 def detect_merge_conflict(argv=None):
     parser = argparse.ArgumentParser()
     parser.add_argument('filenames', nargs='*')
+    parser.add_argument('--assume-in-conflict', action='store_true')
     args = parser.parse_args(argv)
 
-    if not is_in_merge():
+    if not is_in_merge(args):
```
asottile: probably can leave `is_in_merge` alone and just do:

```python
if not is_in_merge() and not args.assume_in_merge:
    return 0
```

(oops, sorry, let's also change the argument to `--assume-in-merge` now that I'm reading the code /o\)

vinayinvicible: fixed

tests/check_merge_conflict_test.py
```diff
@@ -128,5 +128,12 @@ def test_ignores_binary_files():
 
 
 def test_does_not_care_when_not_in_a_merge(tmpdir):
-    tmpdir.join('README.md').write('problem\n=======\n')
-    assert detect_merge_conflict(['README.md']) == 0
+    f = tmpdir.join('README.md')
+    f.write('problem\n=======\n')
+    assert detect_merge_conflict([str(f.realpath())]) == 0
```
asottile: can just use `f.strpath` here instead of `str(f.realpath())`

thanks for fixing this test too 👍

vinayinvicible: `strpath` is not officially [documented](http://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath). this is why I've used `realpath`

asottile: interesting! `str(f)` will work too.  I'll work on getting that documented from the `py` side 😆

asottile:
Thanks @vinayinvicible this is awesome 🎉

vinayinvicible:
Thanks for instant feedbacks :+1:

PR #304, opened 2018-07-05
Added Python 3.7 support

expobrain:


asottile:
(you and I both know so I'm not going to dupe the comments here, but let's discuss in https://github.com/pre-commit/pre-commit/pull/789#pullrequestreview-134841600)

asottile:
Thanks!

PR #308, opened 2018-07-27
Add `forbid-type-hints`

danielmmetz:
Intended for use in Python3.5+ to encourage use of type annotations instead.


apallin:
🐑

asottile:
(for context, we've discussed this offline and we're soon going to be adding a `pygrep-hooks` repo which contains a bunch of these 🎉)

PR #312, opened 2018-09-07
Exclude folders from file list

zbarryte:
**Issue**

If a project contains a git submodule, this submodule may be mistakenly listed as a file, and an `open` operation will then be called on it, resulting in a runtime error.

**Resolution**

This PR addresses that issue by performing an explicit check to exclude folders from the check intended to be performed on files.

### DIFF
--- a/pre_commit_hooks/detect_private_key.py
+++ b/pre_commit_hooks/detect_private_key.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import argparse
+import os
 import sys
 
 BLACKLIST = [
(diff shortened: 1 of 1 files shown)


asottile:
What version of pre-commit are you using? This should not be necessary since this uses `text` type

zbarryte:
`0.13.3`, but I just ran `--upgrade` when prompted to be at least at version `0.15.0`.

asottile:
You're running an extremely old version of pre-commit (and probably of the hooks as well)

The current version of pre-commit is `1.11.0`, you can upgrade the hooks via `pre-commit autoupdate`

zbarryte:
Ah -- thank you!

asottile:
just to validate with the latest versions of things it works for me so I'm going to close this, if you can reproduce with the latest version please feel free to comment here and I can reopen:

```
$ pre-commit --version
pre-commit 1.11.0
t$ cat .pre-commit-config.yaml 
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v1.4.0
    hooks:
    -   id: detect-private-key
$ pre-commit run --all-files
Detect Private Key.......................................................Passed
$ git ls-files
.gitmodules
.pre-commit-config.yaml
pre-commit
$ git submodule foreach git remote -v
Entering 'pre-commit'
origin	git@github.com:pre-commit/pre-commit (fetch)
origin	git@github.com:pre-commit/pre-commit (push)
```

zbarryte:
Sounds good to me!

PR #318, opened 2018-10-04
Fix readme to correctly reflect no-commit-to-branch args

milin:
Current documentation to use the `no-commit-to-branch` is kinda confusing. So updated it to reflect how it correctly works.


README.md
```diff
@@ -82,7 +82,7 @@ Add this to your `.pre-commit-config.yaml`
 - `name-tests-test` - Assert that files in tests/ end in `_test.py`.
     - Use `args: ['--django']` to match `test*.py` instead.
 - `no-commit-to-branch` - Protect specific branches from direct checkins.
-    - Use `args: [--branch <branch>]` to set the branch. `master` is the
+    - Use `args: ['staging', 'master']` to set the branch. `master` is the
```
asottile: ah yeah this would be `args: [--branch, staging, --branch, master]` -- good catch!

PR #330, opened 2018-10-23
Automatically fix package=version in requirements.txt

### LINKED ISSUE #284: requirements-txt-fixer: auto fix = to ==
Currently, the **requirements-txt-fixer** is not fixing this kind of entries: **libraryname=1.0**
The problem is quite simple and happens very often when people forget to add the second **=** sign. Do you think that fixing this **libraryname=1.0** to **libraryname==1.0** would be a great idea? From time to time we have CI tests failing because we didn't catch this typo.

hackedd:
Fixes #284.

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -61,6 +61,9 @@ def fix_requirements(f):
         elif line.startswith(b'#') or line.strip() == b'':
             requirement.comments.append(line)
         else:
+            # Automatically fix 'package=version' to 'package==version'
+            if b'=' in line and b'==' not in line and b'egg=' not in line:
+                line = line.replace(b'=', b'==')
             requirement.value = line
 
     # if a file ends in a comment, preserve it at the end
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -61,6 +61,9 @@ def fix_requirements(f):
         elif line.startswith(b'#') or line.strip() == b'':
             requirement.comments.append(line)
         else:
+            # Automatically fix 'package=version' to 'package==version'
+            if b'=' in line and b'==' not in line and b'egg=' not in line:
```
asottile: it looks like there's a few other cases that aren't covered here.  Reading through https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers these are things that might be valid in a `requirements.txt`:

```
mypy;python_version>="3.5"
foopkg --hash=sha256:12341234123412341234
--index-url=https://example.com/simple
```

(it would be good to add a test for these as well!)

hackedd: Yeah, I came to the same conclusion (and created #331).

asottile:
yeah I suspect this isn't possible or desirable and given we couldn't figure out how to move forward I'm going to close this and the related issue

-- thanks again for the PR though @hackedd -- sorry we never really got it over the line

PR #333, opened 2018-10-28
Fix/aws credentials docs

rpdelaney:
* Documents the previously undocumented `--allow-missing-credentials` switch to `detect-aws-credentials`. 
* Documents the required argument to `--credentials-file`

By the way, `--credentials-file` seems to be a misnomer: it takes _n_ files and even the destination var is named `credential_files`. Maybe adding `--credential-files` and undocumenting `--credentials-file` (for backward compatibility) would make sense. If you agree I can send a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Add this to your `.pre-commit-config.yaml`
 - `detect-aws-credentials` - Checks for the existence of AWS secrets that you
   have set up with the AWS CLI.
   The following arguments are available:
-  - `--credentials-file CREDENTIAL_FILES` - additional AWS CLI style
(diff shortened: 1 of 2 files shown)


rpdelaney:
Ah, that makes more sense. If you can hold  merging this I'll fix that up when I get a chance soon.

asottile:
> Ah, that makes more sense. If you can hold merging this I'll fix that up when I get a chance soon.

will do! -- let me know when you want me to take a look at this again

rpdelaney:
I got some time sooner than I thought. :o) How's that?

asottile:
Thanks again!

PR #337, opened 2018-11-15
Adding detect-raw-datetime-manipulation option

goheer:
This will allow us to detect instances where we are manually manipulating datetime by using timedelta. Also it will flag all instances where we create datetime by manually passing tzinfo.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -105,6 +105,12 @@
     entry: detect-private-key
     language: python
     types: [text]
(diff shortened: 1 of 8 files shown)


goheer:
@matthewgrossman @asottile

goheer:
meow @dan98765

asottile:
> @goheer could you please share some more context on the motivation behind why you want to detect where dt/timedelta are being used like this?

Sorry about this, this was a postmortem action item which was accidentally opened against an open source repository -- we're going to implement this in a different way.

PR #342, opened 2018-11-26
fix(metadata): remove `autopep8-wrapper`

TheKevJames:
`autopep8-wrapper` was moved to `pre-commit/mirrors-autopep8` as of release v2.0.0.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,10 +1,3 @@
--   id: autopep8-wrapper
-    name: autopep8 wrapper
-    description: "Runs autopep8 over python source.  If you configure additional arguments you'll want to at least include -i."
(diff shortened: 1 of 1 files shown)


asottile:
Keeping this is intentional. The hook invariably fails with a useful message telling the consumer how to migrate

TheKevJames:
Ah, ok. Good to know! You should be aware that this means the hook still appears on [this webpage](https://pre-commit.com/hooks.html), then. Maybe its worth adding a deprecation notice to the `description`, as you do with the `pyflakes` check?

asottile:
Ah yeah probably should update the description, want to make a PR for that?

TheKevJames:
[Here ya go](https://github.com/pre-commit/pre-commit-hooks/pull/343)

PR #343, opened 2018-11-26
fix(metadata): deprecate `autopep8-wrapper`

TheKevJames:


.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,6 @@
 -   id: autopep8-wrapper
-    name: autopep8 wrapper
-    description: "Runs autopep8 over python source.  If you configure additional arguments you'll want to at least include -i."
+    name: autopep8 wrapper (DEPRECATED, use pre-commit/mirrors-autopep8)
+    description: "Runs autopep8 over python source.  If you configure additional arguments you'll want to at least include -i. (This is deprecated, use pre-commit/mirrors-autopep8)."
```
asottile: Let's remove the old description, it's not relevant now.

Adjusting the name isn't necessary, let's leave that

.pre-commit-hooks.yaml
```diff
@@ -1,6 +1,6 @@
 -   id: autopep8-wrapper
-    name: autopep8 wrapper (DEPRECATED, use pre-commit/mirrors-autopep8)
-    description: "Runs autopep8 over python source.  If you configure additional arguments you'll want to at least include -i. (This is deprecated, use pre-commit/mirrors-autopep8)."
+    name: autopep8 wrapper
+    description: "This is deprecated, use pre-commit/mirrors-autopep8."
```
asottile: - the quotes do nothing here
- add the word instead at the end of the sentence

asottile: oh I mean insert the word "instead" so it reads:

>  This is deprecated, use pre-commit/mirrors-autopep8 instead.

asottile:
Thanks!~

PR #345, opened 2018-12-12
add a encoding optional for read python code

x007007007:


asottile:
while I personally think that python source should always be UTF-8 encoded, I'm willing to take this patch if you can write a test demonstrating this feature.  Let me know if you'd like some pointers as to where and how to test this :)

asottile:
via #374

PR #347, opened 2018-12-26
Add update_commit_msg

milin:
Implements https://github.com/pre-commit/pre-commit/issues/884

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -183,3 +183,9 @@
     entry: trailing-whitespace-fixer
     language: python
     types: [text]
+-  id: prepend-ticket-number-to-commit-msg
+   name: prepend-ticket-number-to-commit-msg
+   entry: prepend-ticket-number-to-commit-msg
+   language: python
+   stages: [commit-msg]
+   description: prepend-ticket-number-to-commit-msg - Use it to prepend your commits with info from your branch. See Readme for info.
--- a/pre_commit_hooks/prepend_ticket_number_to_commit_msg.py
+++ b/pre_commit_hooks/prepend_ticket_number_to_commit_msg.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import argparse
+import io
+import re
+import subprocess
+import sys
+
+import six
+
+
+def update_commit_message(filename, regex):
+    with io.open(filename, 'r+') as fd:
+        contents = fd.readlines()
+        commit_msg = contents[0]
+        # Check if we can grab ticket info from branch name.
+        branch = get_branch_name()
+        if all([
+            not re.search(regex, commit_msg),
+            re.search(regex, branch),
+        ]):
+            ticket = branch.split(six.text_type('_'))[0]
+            new_commit_msg = '{} {}'.format(ticket, commit_msg)
+            fd.seek(0)
+            fd.write(
+                six.text_type(
+                    new_commit_msg,
+                ),
+            )
+            fd.truncate()
+
+
+def get_branch_name():
+    # Only git support for right now.
+    return subprocess.check_output(
+        [
+            'git',
+            'rev-parse',
+            '--abbrev-ref',
+            'HEAD',
+        ],
+    ).decode('UTF-8')
+
+
+def main(argv=None):
+    """This hook saves developers time by prepending ticket numbers to commit-msgs.
+    For this to work the following two conditions must be met:
+        - The ticket format regex specified must match.
+        - The branch name format must be <ticket number>_<rest of the branch name>
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='+')
+    parser.add_argument('--regex')
+    args = parser.parse_args(argv)
+    regex = args.regex or '[A-Z]+-\d+'  # noqa
+    update_commit_message(args.filenames[0], regex)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ Add this to your `.pre-commit-config.yaml`
       use `args: [--markdown-linebreak-ext=md]` (or other extensions used
       by your markdownfiles).  If for some reason you want to treat all files
       as markdown, use `--markdown-linebreak-ext=*`.
+- `prepend-ticket-number-to-commit-msg` - Use it to prepend your commits with info from your branch, like ticket numbers. For e.g. if you name your branch `JIRA-1234_awesome_feature` and commit `Fix some bug`, the commit will be updated to `JIRA-1234 Fix some bug`. Pass `--regex=` or update `args: [--regex=<custom regex>]` in your .yaml file if you have custom ticket regex. By default its `[A-Z]+-\d+`.
 
 ### Deprecated / replaced hooks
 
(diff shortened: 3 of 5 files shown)


.pre-commit-hooks.yaml
```diff
@@ -183,3 +183,13 @@
...
     types: [text]
+-  id: update-commit-msg
+   name: update-commit-msg
+   entry: update-commit-msg
+   # Update to your ticket matching regex
+   args: ['--regex=TICKET-[0-9]']
+   language: python
+   stages: [commit-msg]
+   additional_dependencies: ['six', 'invoke']
+   files: .git/COMMIT_EDITMSG
```
asottile: this is unnecessary (and in some cases incorrect) -- pre-commit will make sure to handle passing the right filenames here.  In the latest version, neither `types` nor `files` is required so it'll just do the right thing

.pre-commit-hooks.yaml
```diff
@@ -183,3 +183,13 @@
...
     language: python
     types: [text]
+-  id: update-commit-msg
+   name: update-commit-msg
+   entry: update-commit-msg
+   # Update to your ticket matching regex
+   args: ['--regex=TICKET-[0-9]']
+   language: python
+   stages: [commit-msg]
+   additional_dependencies: ['six', 'invoke']
```
asottile: these should be specified in setup.py, six is already there, invoke is (imo) unnecessary because you can just use `subprocess.check_output`

.pre-commit-hooks.yaml
```diff
@@ -183,3 +183,13 @@
     entry: trailing-whitespace-fixer
     language: python
     types: [text]
+-  id: update-commit-msg
+   name: update-commit-msg
+   entry: update-commit-msg
+   # Update to your ticket matching regex
+   args: ['--regex=TICKET-[0-9]']
```
asottile: I'd probably just default this inside the hook itself, and leave out `args` here.

I'd probably default to `[A-Z]+-\d+`

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, print_function, unicode_literals
+import sys
+import optparse
```
asottile: optparse should be considered deprecated in python, use argparse instead

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+from invoke import run
+
+import re
+
+
+class UpdateCommitMsg(object):
```
asottile: this is unnecessarily object-oriented.  A better approach (imo) is to have a main function which calls out to the check_file function.  All of the other hooks in this repository are written in that way so you can find a few examples.  My bare minimum setup usually looks like this:

```python
import argparse

...

def main(argv=None):
    parser.add_argument('filenames', nargs='+')

    ...

if __name__ == '__main__':
    exit(main())
```

this allows you to have a testable interface (`parser.parse_args(None)` will fall back to sys.argv)

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    This hook saves developers time by prepending ticket numbers to commit-msgs.
```
asottile: v v minor, docstrings shouldn't start with whitespace, (and single-line shouldn't end with whitespace)

```python
"""yes"""

"""yes

yes
"""

"""no
"""

"""
no
no
"""
```

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+                if re.search(regex, self.branch):
+                    ticket = self.branch.split('_')[0]
+                    new_commit_msg = '{} {}'.format(ticket, commit_msg)
+                    fd.seek(0)
+                    fd.write(new_commit_msg)
+                    fd.truncate()
+
+    def get_branch_name(self):
+        # Only git support for right now.
+        return run("git rev-parse --abbrev-ref HEAD").stdout
```
asottile:
```
subprocess.check_output(('git', 'rev-parse', '--abbref-ref', 'HEAD')).decode('UTF-8')
```

there's also potentially a failure mode when you're not on a branch that might be worth handling or ignoring (?)

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+            description='Automate prepending ticket info to your commits.'
+        )
+        parser.add_option(
+            "-r",
+            "--regex",
+            help="write report to FILE",
+        )
+        (opts, files) = parser.parse_args()
+
+        regex = r'{}'.format(opts.regex)
```
asottile: this line does nothing (the `r` doesn't change how the string is treated, and `'{}'.format(x)` is the same as `str(x)` and `str(a_str)` is the same as `a_str`)

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+        if len(files) == 0:
+            parser.error('No filenames provided')
```
asottile: at least from pre-commit's perspective, 0 filenames isn't an error it's just the base case.  that said, pre-commit will never call a hook without filenames (unless you explicitly ask for `pass_filenames: false` and/or `always_run: true`)

pre_commit_hooks/update_commit_msg.py
```diff
@@ -0,0 +1,65 @@
...
+        - The ticket format regex specified must match.
+        - The branch name format must be <ticket number>_<rest of the branch name>
+    """
+
+    def __init__(self):
+        self.errors = 0
+        self.branch = self.get_branch_name()
+
+    def check_file(self, filename, regex):
+        self.current_filename = filename
```
asottile: assigned but never used, also assigned outside of `__init__` is often a smell

.pre-commit-hooks.yaml
milind-shakya-sp: I think the reason why i added it is, without it, it was having some side-effects and it was updating files all changed files.

pre_commit_hooks/update_commit_msg.py
milind-shakya-sp: Sounds fair.

.pre-commit-hooks.yaml
asottile: hmm, probably before you added `stages: [commit-msg]`?

milind-shakya-sp: yeah probably, i cannot reproduce it now. will remove.

pre_commit_hooks/prepend_ticket_number_to_commit_msg.py
```diff
@@ -0,0 +1,63 @@
...
+    """This hook saves developers time by prepending ticket numbers to commit-msgs.
+    For this to work the following two conditions must be met:
+        - The ticket format regex specified must match.
+        - The branch name format must be <ticket number>_<rest of the branch name>
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='+')
+    parser.add_argument('--regex')
+    args = parser.parse_args(argv)
+    regex = args.regex or '[A-Z]+-\d+'  # noqa
```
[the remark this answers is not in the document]
asottile: I'm going to handle that part here: https://github.com/pre-commit/pre-commit-hooks/pull/352

milin:
@asottile Thanks for the review. I am closing this pr, and it will instead be its own repo. Thanks for the suggestion.

asottile:
no problem! when you're ready we can add it to the official list https://github.com/pre-commit/pre-commit.github.io (editing all-repos.yaml)

PR #351, opened 2018-12-28
switch from pyyaml to ruamel.yaml

### LINKED ISSUE #350: Switch from pyyaml -> ruamel.yaml
Seems better maintained, I've asked about the "new api" and it seems like there'll be some stability there: https://bitbucket.org/ruamel/yaml/issues/267/id-like-to-use-the-new-api

Should be pretty easy to swap

At the very least we get duplicate key checking by default

asottile:
Resolves #350

### DIFF
--- a/pre_commit_hooks/check_yaml.py
+++ b/pre_commit_hooks/check_yaml.py
@@ -4,12 +4,9 @@ import argparse
 import collections
 import sys
 
-import yaml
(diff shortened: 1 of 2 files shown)


asottile:
CC @ssbarnea -- this will enable duplicate key checking in `check-yaml`

PR #365, opened 2019-02-16
Add hook for checking python modules

jcomo:
Adds a simple hook for checking that python modules have their `__init__.py` file. This is something I've been using in local development, and thought it would be useful to have available more broadly; let me know what you think @asottile


asottile:
It'll soon be obsolete once python2.7 dies and PEP420 can be used more widely :thinking:

jcomo:
Yup, sounds good. Closing

Issue #309, opened 2018-08-01
Enable requirements-txt-fixer to sort and fix  dependency list in Pipfile


Motor-taxi-master-rider:
Since **pipenv** is getting more popularity, a bunch of python libs try to store their dependencies into a **Pipfile**. Currently, our pre-commit supports _requirements-txt-fixer_ for the **reqirement.txt**. I think we should support the similar function in the  _requirements-txt-fixer_ for the **Pipfile** guys.

**Pipfile** is kind of a _TOML_ file, we may have a _TOML_ sorter and formatter in this Pipfile fixer or something.

asottile:
The unfortunate thing about `Pipfile`s is they're a bit more free form.  There isn't necessarily a good overall sorting for them.

Additionally, I have yet to find a round-trip `toml` rewriter.

Everything was just so much simpler with `requirements.txt`.  Would be open to a PR if you can make it work though :)

asottile:
The other odd thing is `pipenv` isn't really designed for libraries -- more so for _applications_ but agree with everything else you've said :)

Motor-taxi-master-rider:
Yes! I agree with you that `pipenv` is designed for _applications_.

For the `Pipfile` part, I've got some rough idea about the implementation. If it is not suitable for this scenario, please feel free to point out my fault. :)

I think we could utilize the python [toml](https://github.com/uiri/toml) module to parse `Pipfile`:
```
with open('Pipfile') as f:
    pipfile=toml.load(f, _dict=OrderedDict)  ## we could use normal dict if cpython version>= 3.6
```

Then we could get a nested ordered dict like:
```
{'source': [{'url': 'https://pypi.org/simple', 'verify_ssl': True, 'name': 'pypi'}], 'packages': {'click': '*', 'flask-mongoengine': '*', 'itsdangerous': '*', 'mongomock': '*', 'python-dotenv': '*'}, 'dev-packages': {'isort': '*', 'pre_commit': '*', 'yapf': '*', 'Faker': '*', 'pytest': '*'}, 'requires': {'python_version': '3.6'}}
```

We may iterate the child dictionary to sort these package names and write back to the `Pipfile` use the `dumps` function.

It is a simple implementation which will add dependency to `pre-commit-hooks`, what do you think? By the way, I am glad to open a PR request to add the features.

asottile:
The main issue is `Pipfile` is a human writable format where comments are important.  The `toml` module doesn't preserve them unfortunately

[...]

PR #373, opened 2019-03-27
Add check-toml hook

MarSoft:
It is important to have such hook
to avoid committing non-valid `pyproject.toml` or `Pipfile`.

Related to #309.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ packages = find:
 install_requires =
     flake8
     ruamel.yaml>=0.15
+    pytoml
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -26,6 +26,7 @@ packages = find:
 install_requires =
     flake8
     ruamel.yaml>=0.15
+    toml
```
asottile: (very minor): alphabetize this list

asottile:
via #400 -- thanks for the PR!

Issue #375, opened 2019-04-09
Allow no-commit-to-branch to block commits based on regex matching


marcjay:
Hopefully not alone in wanting this functionality - our use case is we have multiple release branches (e.g. `release/1.0.0`, `release/2.0.0`) and no commits should be made to these branches directly, only via pull requests.

This would be to add wildcard matching, so that with `args: [--branch, master, --branch, release/*]` it would cause pre-commit to fail if a commit is attempted on any of: `master`, `release/1.0.0` or `release/2.0.0`

asottile:
Seems fine, let's use `--pattern` as a separate argument and use regular expressions instead of glob / fnmatch

marcjay:
Sure, I'll re-work the PR. Just want to double check - do you see the example above becoming:
```
args: [--branch, master, --pattern, release/.*]
```
i.e. the branch and pattern args would become an OR match, as opposed to the two args being mutually exclusive and needing to do:
```
args: [--pattern, master|(release/.*)]
```

asottile:
the first one sounds good 👍

asottile:
via #376

PR #376, opened 2019-04-09
Add regex matching to no-commit-to-branch hook

marcjay:
Add wildcard matching to no-commit-to-branch hook, so that commits can be blocked on, for example, all release branches with 'release/*'

Addresses: #375

### DIFF
--- a/pre_commit_hooks/no_commit_to_branch.py
+++ b/pre_commit_hooks/no_commit_to_branch.py
@@ -2,16 +2,16 @@ from __future__ import print_function
 
 import argparse
 import re
+from typing import FrozenSet
 from typing import Optional
 from typing import Sequence
-from typing import Set
 
 from pre_commit_hooks.util import CalledProcessError
 from pre_commit_hooks.util import cmd_output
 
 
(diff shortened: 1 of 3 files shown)


pre_commit_hooks/no_commit_to_branch.py
```diff
@@ -10,14 +10,17 @@
 from pre_commit_hooks.util import cmd_output
 
 
-def is_on_branch(protected):  # type: (Set[str]) -> bool
+def is_on_branch(protected, patterns=set()):
```
asottile: the default value here should be immutable (such as `frozenset()`)

marcjay: Cheers for the review :smile:! Have made the change. Also had to change the calling function to satisfy the type checker

asottile:
cool! I switched it to `AbstractSet` to appease mypy

marcjay:
Thanks! Appreciate the crash-course in mypy and cheers for fixing

asottile:
And thank you for the pull request 🎉

PR #399, opened 2019-08-03
Adds remove-tabs hook

ssbarnea:
Based on original work from https://github.com/Lucas-C/pre-commit-hooks

### DIFF
--- a/pre_commit_hooks/remove_tabs.py
+++ b/pre_commit_hooks/remove_tabs.py
@@ -1,0 +1,53 @@
+from __future__ import print_function
+
+import argparse
+import sys
+from typing import Any
+
+
+def contains_tabs(filename):  # type: (str) -> bool
+    with open(filename, mode='rb') as file_checked:
+        return b'\t' in file_checked.read()
+
+
+def removes_tabs_in_file(filename, whitespaces_count):
+    # type: (str, int) -> None
+    with open(filename, mode='rb') as file_processed:
+        lines = file_processed.readlines()
+    lines = [line.replace(b'\t', b' ' * whitespaces_count) for line in lines]
+    with open(filename, mode='wb') as file_processed:
+        for line in lines:
+            file_processed.write(line)
+
+
+def main(argv=None):  # type: (Any) -> int
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--whitespaces-count', type=int, required=True,
+        help='number of whitespaces to substitute tabs with',
+    )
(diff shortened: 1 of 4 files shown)


pre_commit_hooks/util.py
```diff
@@ -7,6 +7,9 @@
 from typing import Set
 
 
+KNOWN_BINARY_FILE_EXTS = ('.pdf',)
```
asottile: this shouldn't be necessary, `identify` will filter out binary files if you use `types: [text]`

pre_commit_hooks/remove_tabs.py
```diff
@@ -0,0 +1,52 @@
...
+        print(
+            'You can check the changes made. Then simply '
+            '"git add --update ." and re-commit',
+        )
+        return 1
+    return 0
+
+
+if __name__ == '__main__':
+    sys.exit(main(sys.argv[1:]))
```
asottile: can you follow the pattern used in the other hooks?

Lucas-C:
@ssbarnea @asottile : the code license stays the same and I am totally fine with integrating hooks from Lucas-C/pre-commit-hooks here. I mostly agree with your comments in https://github.com/Lucas-C/pre-commit-hooks/issues/11 @ssbarnea about the benefit of having hooks like this one, with no dependency to a specific language/tool, grouped in a single repo.
It will mean moving the very light support / maintenance for this hook to the maintainers of this repo, and I'm cool with that as long as @asottile & the other admins are OK.

Btw, is there an officiel position of the `pre-commit` core devs about this ? Do you recommend contributing to this repo instead of isolated projects ?

ssbarnea:
@Lucas-C Thanks! I guess it makes sense to consolidate generic mini-hooks, it makes much easier to adopt (and maintain) them.

asottile:
I've had some time to think about this and while this particular hook fits the category for this repository (high quality, language-agnostic, pretty useful) I think that I'd rather celebrate @Lucas-C's contributions and continue to direct people to their many pre-commit hook repositories than cherry pick individual hooks here.  In particular for this hook, it really wouldn't make sense to take it without the three+ related hooks and at that point it would really be merging another repository into this one which I don't really want to do.  It's easy / trivial to add another repo to your .pre-commit-config.yaml and so I don't see the points made in https://github.com/Lucas-C/pre-commit-hooks/issues/11 being valid.

Thanks again for the PR @ssbarnea but please in the future ask nicely of the author (and CC me and/or @chriskuehl) before uprooting and moving code around and especially don't do so without LICENSE / attribution

PR #400, opened 2019-08-03
Toml checker

ssbarnea:
This should enable us to validate TOML files, like the controversial `pyproject.toml` ones.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -70,6 +70,12 @@
     entry: check-symlinks
     language: python
     types: [symlink]
+-   id: check-toml
+    name: Check Toml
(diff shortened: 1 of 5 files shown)


testing/resources/toml_bad.toml
```diff
@@ -0,0 +1,3 @@
+key = # INVALID
+
+= "no key name"  # INVALID
```
asottile: instead of having invalid toml files checked in, can you write them as temporary files in the tests -- I realize there's old hooks which do the former but I've been preferring the latter for everything new -- thanks!

ssbarnea:
@asottile regarding your worries around which toml to use. I have the impression that there is only one currently maintained: https://github.com/uiri/toml

Issue #402, opened 2019-08-03
git lfs install fails when run from inside tox when HOME is not defined


ssbarnea:
CI is currently broken as we can see failures unrelated to our patches:
```
cmd = ('git', 'lfs', 'install'), kwargs = {'stderr': -1, 'stdout': -1}
retcode = 0, proc = <subprocess.Popen object at 0x7f33a741fbd0>
stdout = "WARNING: Error running /usr/lib/git-core/git 'config' '--global' '--replace-all' 'filter.lfs.required' 'true': 'fatal: $HOME not set' 'exit status 128'
Run `git lfs install --force` to reset git config.
"
stderr = ''

    def cmd_output(*cmd, **kwargs):  # type: (*str, **Any) -> str
        retcode = kwargs.pop('retcode', 0)
        kwargs.setdefault('stdout', subprocess.PIPE)
        kwargs.setdefault('stderr', subprocess.PIPE)
        proc = subprocess.Popen(cmd, **kwargs)
        stdout, stderr = proc.communicate()
        stdout = stdout.decode('UTF-8')
        if retcode is not None and proc.returncode != retcode:
>           raise CalledProcessError(cmd, retcode, proc.returncode, stdout, stderr)
E           CalledProcessError: ((u'git', u'lfs', u'install'), 0, 2, u"WARNING: Error running /usr/lib/git-core/git 'config' '--global' '--replace-all' 'filter.lfs.required' 'true': 'fatal: $HOME not set' 'exit status 128'\nRun `git lfs install --force` to reset git config.\n", '')

```

asottile:
interesting, I wonder what changed in git-lfs 🤔

ssbarnea:
I never used git-lfs but here is what found so far: problem is not related to Windows. I was able to reproduce it on my MacOS box easily by installing git-lfs. Once installed the test suite will stop skipp testing it.

When running `git lfs install` outside tox, you get a a sucess. If you run it from inside tox you get:
```
WARNING: Error running /usr/local/Cellar/git/2.22.0_1/libexec/git-core/git 'config' '--global' '--replace-all' 'filter.lfs.clean' 'git-lfs clean -- %f': 'fatal: $HOME not set' 'exit status 128'
Run `git lfs install --force` to reset git config.
ERROR: InvocationError for command /usr/local/bin/git lfs install (exited with code 2)
```

PR #403, opened 2019-08-03
Assure we have HOME defined during testing

ssbarnea:
Fixed broken CI caused by missing HOME.

Fixes: #402

### DIFF
--- a/tests/check_added_large_files_test.py
+++ b/tests/check_added_large_files_test.py
@@ -76,8 +76,9 @@ xfailif_no_gitlfs = pytest.mark.xfail(
 
 
 @xfailif_no_gitlfs
-def test_allows_gitlfs(temp_git_dir):  # pragma: no cover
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -8,6 +8,8 @@ setenv =
     GIT_COMMITTER_NAME = "test"
     GIT_AUTHOR_EMAIL = "test@example.com"
     GIT_COMMITTER_EMAIL = "test@example.com"
+    # git-lfs tests need HOME because git itself needs one
+    HOME={homedir)
```
asottile: this isn't ~quite right as `HOME` means nothing on windows -- does adding `HOME` to `passenv` fix this?

asottile: actually, now that I think about it, let's just set HOME for the tests in question and point it at a temporary directory -- that way the test suite doesn't modify global settings

asottile:
thanks for the patch! 🎉

PR #408, opened 2019-08-15
Add logic to print diff between expected and actual JSON

joepin:
This PR makes the pretty JSON check more verbose when it encounters errors, that way developers can see which lines are causing errors in order to debug.

I haven't seen any talk about adding this anywhere, but would love to hear the maintainers' thoughts on this.

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -4,6 +4,7 @@ import argparse
 import io
 import json
 import sys
+import difflib
 from collections import OrderedDict
 from typing import List
 from typing import Mapping
@@ -55,6 +56,14 @@ def parse_topkeys(s):  # type: (str) -> List[str]
     return s.split(',')
 
 
+def get_diff(source, target):
+    source_lines = ''.join(source).split('\n')
+    target_lines = ''.join(target).split('\n')
+    d = difflib.Differ()
+    diff = d.compare(source_lines, target_lines)
+    return '\n'.join(diff)
+
+
 def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
     parser = argparse.ArgumentParser()
     parser.add_argument(
@@ -96,6 +105,13 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
         default=[],
         help='Ordered list of keys to keep at the top of JSON hashes',
     )
+    parser.add_argument(
+        '--show-expected',
+        action='store_true',
+        dest='show_expected',
+        default=False,
+        help='Show a diff between the input file and expected (pretty) output',
+    )
 
     parser.add_argument('filenames', nargs='*', help='Filenames to fix')
     args = parser.parse_args(argv)
@@ -115,16 +131,8 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             if contents != pretty_contents:
                 print('File {} is not pretty-formatted'.format(json_file))
 
-                contents_by_line = ''.join(contents).split('\n')
-                pretty_contents_by_line = ''.join(pretty_contents).split('\n')
-
-                diff = len(contents_by_line) - len(pretty_contents_by_line)
-                if diff > 0:
-                    pretty_contents_by_line.extend([''] * diff)
-
-                for line_num, line in enumerate(contents_by_line):
-                    if line != pretty_contents_by_line[line_num]:
-                        print('{}:{}'.format(json_file, line_num))
+                if args.show_expected:
+                    print(get_diff(contents, pretty_contents))
 
                 if args.autofix:
                     _autofix(json_file, pretty_contents)
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -115,6 +115,17 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             if contents != pretty_contents:
                 print('File {} is not pretty-formatted'.format(json_file))
 
+                contents_by_line = ''.join(contents).split('\n')
+                pretty_contents_by_line = ''.join(pretty_contents).split('\n')
+
+                diff = len(contents_by_line) - len(pretty_contents_by_line)
+                if diff > 0:
+                    pretty_contents_by_line.extend([''] * diff)
+
+                for line_num, line in enumerate(contents_by_line):
+                    if line != pretty_contents_by_line[line_num]:
+                        print('{}:{}'.format(json_file, line_num))
+
                 if args.autofix:
                     _autofix(json_file, pretty_contents)
 
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -120,7 +120,10 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             )
 
             if contents != pretty_contents:
-                print('File {} is not pretty-formatted'.format(json_file))
+                print(
+                    'File {} is not pretty-formatted'.format(json_file),
+                    file=sys.stderr,
+                )
 
                 if args.autofix:
                     _autofix(json_file, pretty_contents)
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -56,12 +56,12 @@ def parse_topkeys(s):  # type: (str) -> List[str]
     return s.split(',')
 
 
-def get_diff(source, target):  # type: (List[str], List[str]) -> str
-    source_lines = ''.join(source).split('\n')
-    target_lines = ''.join(target).split('\n')
-    d = difflib.Differ()
-    diff = d.compare(source_lines, target_lines)
-    return '\n'.join(diff)
+def get_diff(source, target):  # type: (str, str) -> str
+    source_lines = source.splitlines(True)
+    target_lines = target.splitlines(True)
+    diff = ''.join(difflib.ndiff(source_lines, target_lines))
+    print(diff)
+    return diff
 
 
 def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/pretty_format_json.py
```diff
@@ -115,6 +115,17 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
...
+                contents_by_line = ''.join(contents).split('\n')
+                pretty_contents_by_line = ''.join(pretty_contents).split('\n')
+
+                diff = len(contents_by_line) - len(pretty_contents_by_line)
+                if diff > 0:
+                    pretty_contents_by_line.extend([''] * diff)
+
+                for line_num, line in enumerate(contents_by_line):
+                    if line != pretty_contents_by_line[line_num]:
+                        print('{}:{}'.format(json_file, line_num))
```
asottile: probably just use `difflib` to print a diff instead

could you add a test as well?

joepin: Any advice on how to test? From what I can see, these changes don't change the return value of the function.

Do you think it would be more beneficial to put a diff behind a flag? That way it's more easily tested, and it doesn't alter default behavior?

asottile: You can use the [`capsys`](https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function) pytest fixture -- there's a few examples in some of the tests already iirc

I would only print the diff in the case of the non-auto-fixing mode (so it's already behind a flag)

joepin: Ok, will look in a bit - thanks!

pre_commit_hooks/pretty_format_json.py
```diff
@@ -55,6 +56,14 @@ def parse_topkeys(s):  # type: (str) -> List[str]
     return s.split(',')
 
 
+def get_diff(source, target):
+    source_lines = ''.join(source).split('\n')
+    target_lines = ''.join(target).split('\n')
```
asottile: shouldn't need `''.join(...)` here, also you'll want to use `.splitlines(True)` and then `''.join(...)` later

pre_commit_hooks/pretty_format_json.py
```diff
@@ -55,6 +56,14 @@ def parse_topkeys(s):  # type: (str) -> List[str]
...
+    d = difflib.Differ()
+    diff = d.compare(source_lines, target_lines)
```
asottile: there's a shortcut to this, you can do `''.join(difflib.ndiff(source_lines, target_lines))`

pre_commit_hooks/pretty_format_json.py
```diff
@@ -115,6 +131,9 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             if contents != pretty_contents:
                 print('File {} is not pretty-formatted'.format(json_file))
 
+                if args.show_expected:
+                    print(get_diff(contents, list(pretty_contents)))
```
asottile: I don't think we need a new option, can just do this in the `else` condition of `if args.autofix:` probably

joepin: Ah yeah, I didn’t see your earlier comments before this change - I’ll switch it out and make the code changes above when I get a chance!

joepin: How does the test function look?

asottile: A couple of things:

- you probably don't need to test how `difflib` works
- for table tests in pytest you'd use `pytest.mark.parametrize` (this will make them separate tests)
- I would just do a single test of `main` to demonstrate the behaviour

tests/pretty_format_json_test.py
```diff
@@ -108,76 +107,51 @@ def test_badfile_main():
...
-    3
-  ],
-  "key1": "val1",
-  "key2": 2,
-  "object": {
-    "bool_key": true
+def test_diffing_output(capsys):
+    resource_path = get_resource_path('not_pretty_formatted_json.json')
+    expected_retval = 1
+    expected_diff = '''
```
asottile: looks like this is printing twice

joepin:
@asottile thanks for all the help! Please let me know if my most recent change looks good to you 🙂

tests/pretty_format_json_test.py
joepin: Good catch! I had a weird print statement in `get_diff` function; fixed in 93b7b66

asottile:
let's print the filename stuff to stderr (and flush)

joepin:
> let's print the filename stuff to stderr (and flush)

I'm not sure what you mean by that - do you mean to say [this](https://github.com/pre-commit/pre-commit-hooks/blob/0408045bd1118e706f789f2611b04f211d962bb7/pre_commit_hooks/pretty_format_json.py#L116) should be printed to stderr? (As it stands now, the main func prints both the file name thing and the diff output to stdout)

asottile:
> 
> 
> > let's print the filename stuff to stderr (and flush)
> 
> I'm not sure what you mean by that - do you mean to say [this](https://github.com/pre-commit/pre-commit-hooks/blob/0408045bd1118e706f789f2611b04f211d962bb7/pre_commit_hooks/pretty_format_json.py#L116) should be printed to stderr? (As it stands now, the main func prints both the file name thing and the diff output to stdout)

yes

pre_commit_hooks/pretty_format_json.py
```diff
@@ -120,7 +120,10 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             )
 
             if contents != pretty_contents:
-                print('File {} is not pretty-formatted'.format(json_file))
+                print(
+                    'File {} is not pretty-formatted'.format(json_file),
+                    file=sys.stderr,
+                )
```
asottile: you'll want a `sys.stderr.flush()` after this as it won't line up always otherwise when piped

tests/pretty_format_json_test.py
```diff
@@ -108,10 +108,8 @@ def test_badfile_main():
 
 
 def test_diffing_output(capsys):
-    resource_path = get_resource_path('not_pretty_formatted_json.json')
     expected_retval = 1
-    expected_diff = '''
-  {
+    expected_out = '''  {
```
asottile: this seems weird to me that this is indented by two spaces instead of one -- just want to double check this

I'd also write this like:
```python
'''\
  {
...
'''
```

just so it's more clear what the actual string renders as

tests/pretty_format_json_test.py
```diff
@@ -126,17 +124,10 @@ def test_diffing_output(capsys):
...
-    actual_retval = main([resource_path])
-    actual_output = capsys.readouterr()
+    actual_retval = main([get_resource_path('not_pretty_formatted_json.json')])
+    out, err = capsys.readouterr()
 
     assert actual_retval == expected_retval
-
-    actual_output = '\n'.join(actual_output)
-    assert actual_output == expected_output
+    assert out == expected_out
```
asottile: should test stderr as well

tests/pretty_format_json_test.py
joepin: Yeah I thought that was weird too, seems to be what difflib prints though so I went with it

asottile: ahhhhh! it should be `difflib.unified_diff` I think!  my bad for suggesting `ndiff` 👍

joepin: No worries! Problem is unified_diff outputs two lines that have trailing whitespace, which flake8 doesn't like. In my latest commit, I was forced to make the expected output string a little less readable in order for flake8 to pass the pre commit check - lmk what you think about that.

tests/pretty_format_json_test.py
```diff
@@ -105,3 +105,33 @@ def test_top_sorted_get_pretty_format():
 def test_badfile_main():
     ret = main([get_resource_path('ok_yaml.yaml')])
     assert ret == 1
+
+
+def test_diffing_output(capsys):
+    resource_path = get_resource_path('not_pretty_formatted_json.json')
+    expected_retval = 1
+    expected_out = '''\
+--- \n+++ \n@@ -1,6 +1,9 @@
```
joepin: This is the line I was talking about. `unified_diff` prints:
```
--- 
+++ 
@@ -1,6 +1,9 @@
```
where the first two lines have a trailing space. This was my temporary solution to pass the test cases and satisfy the pre commit check

asottile: ohh this is where the filename is supposed to be


maybe we should pass in `a/filename` and `b/filename` so that it prints better?

asottile: where `filename` is the actual filename -- so `os.path.join('a', filename)` and `os.path.join('b', filename)`

joepin: Ahh that makes sense - maybe we can just put in some pseudo-names? That way it's a little clearer what the output is saying

So it would be
```python
unified_diff(source, target, fromfile='not_pretty.json', tofile='pretty.json')
```

asottile: yeah, here's an example: https://github.com/asottile/reorder_python_imports/blob/df5d5d28abcf8705bd0c2a622480f6099ca56add/reorder_python_imports.py#L511-L515

joepin: Gotcha - adding it!

pre_commit_hooks/pretty_format_json.py
```diff
@@ -113,10 +120,22 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
...
+                sys.stderr.flush()
 
                 if args.autofix:
                     _autofix(json_file, pretty_contents)
+                else:
+                    print(
+                        get_diff(
+                            ''.join(contents),
```
asottile: this should just be `contents` right?

joepin: 🤔 I remember needing to join it before 

I guess not now that we're using `splitlines`?

tests/pretty_format_json_test.py
```diff
@@ -105,3 +106,37 @@ def test_top_sorted_get_pretty_format():
...
+-  "blah": null
++  "alist": [
++    2,
++    34,
++    234
++  ],
++  "blah": null,
++  "foo": "bar"
+ }}
+
```
asottile: ugh sorry for one more

this extra newline indicates that it should be `print(..., end='')` for the diff (the diff itself includes the ending newline already)

joepin: No worries - updated!

joepin:
Awesome! Thanks for all your help and guidance on this one! 😄

asottile:

PR #413, opened 2019-09-24
Capitalize acronym names

pradyunsg:
I feel a little bad inside every time I look at these names. I finally decided to do something about it.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -48,7 +48,7 @@
     stages: [commit, push, manual]
 -   id: check-json
     name: Check JSON
-    description: This hook checks json files for parseable syntax.
+    description: This hook checks JSON files for parseable syntax.
     entry: check-json
     language: python
     types: [json]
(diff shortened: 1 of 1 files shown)


asottile:
stylistically I actually want to go the other way and remove all the capitalization from the hook names -- would you like to do that instead? (this better matches the file extension and hook ids)

pradyunsg:
I *strongly* prefer all-caps, since they are acronyms and should be capitalized in human-readable strings.

I think we should use all-caps for consistency with any other textual uses, instead of for-the-computer uses -- The [JSON spec](http://json.org/) uses "JSON" and the [TOML spec](https://github.com/toml-lang/toml/blob/548078574af52e79b5e172718df46de852a9bab3/README.md) uses "TOML", in their sentences. I'm pretty sure XML, YAML and others would also be doing the same.

---

I understand the consistency suggestion you make and that's definitely better than status quo. I'm happy to make that change, if you insist on not doing all caps, and I'll update the PR. Just let me know if my paragraph above persuaded you. :)

asottile:
I'm willing to admit I'm wrong, but I'd still like them un-capitalized

while you're at it, I've been meaning to change the `name` for `name-tests-test` to `enforce test file naming` (to make it less confusing when the `--django` flag is passed) -- if you could make that fix as well as part of this that would be great :)

pradyunsg:
I think I'm gonna drop this, since I don't really have much time anytime soon to pick this up again.

I still feel the same as I did when I opened the PR -- the acronyms should be consistently capitalized.

PR #415, opened 2019-09-28
add new line after reading requirements file

### LINKED ISSUE #414: requirements-txt-fixer corrupts file if no new line at end
a file with
```
a
b
c
```
(no new line)
will get "fixed" to:
```
a
bc

```

I was able to debug and identify the issue, I'll be happy to also PR a fix but I was wondering if there is really a use for it? I've not seen any input without a new line at the end in the tests cases and as a work around I can add `end-of-file-fixer` before it.

barakreif:
Resolves #414 
- add new line post reading requirements file.
- change `before` variable from tuple .
- to list, add test cases.

This maintains all previous behaviour(s) as given by the test cases.

(my first ever OS contribution, any and **all**   feedback is welcome)

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -40,8 +40,13 @@ class Requirement(object):
 
 def fix_requirements(f):  # type: (IO[bytes]) -> int
     requirements = []  # type: List[Requirement]
-    before = tuple(f)
+    before = list(f)  # type: List[bytes]
     after = []  # type: List[bytes]
+
+    # adds new line in case one is missing
+    # AND a change to the requirements file is needed regardless:
+    if before and not before[-1].endswith(b'\n'):
+        before[-1] += b'\n'
 
     before_string = b''.join(before)
 
@@ -94,7 +99,7 @@ def fix_requirements(f):  # type: (IO[bytes]) -> int
 
     after_string = b''.join(after)
 
-    if before_string == after_string:
+    if before_string.rstrip() == after_string.rstrip():
         return PASS
     else:
         f.seek(0)
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -43,12 +43,12 @@ def fix_requirements(f):  # type: (IO[bytes]) -> int
     before = list(f)  # type: List[bytes]
     after = []  # type: List[bytes]
 
+    before_string = b''.join(before)
+
     # adds new line in case one is missing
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -94,7 +99,7 @@ def fix_requirements(f):  # type: (IO[bytes]) -> int
 
     after_string = b''.join(after)
 
-    if before_string == after_string:
+    if before_string.rstrip() == after_string.rstrip():
```
asottile: I would just leave this line alone personally -- I think adding a newline to the end of the file is a good fix in the general case 👍

barakreif: will do, but as a code design question, why not to run the `end_of_file_fixer` at first as is?

asottile: Some prefer to pick and choose hooks, I agree someone should probably be configuring the end-of-file-fixer as one of their default hooks -- but it's probably ok to have the behaviour happen here as well

pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -40,11 +40,16 @@ def __lt__(self, requirement):  # type: (Requirement) -> int
 
 def fix_requirements(f):  # type: (IO[bytes]) -> int
     requirements = []  # type: List[Requirement]
-    before = tuple(f)
+    before = list(f)  # type: List[bytes]
```
asottile: just an aside, the type annotation isn't required here -- mypy will properly infer that this is a list of `bytes` objects so generally you can leave it out -- I usually prefer to only type annotate things when mypy requires it (the inference often saves some keyboard-typing and importing)

barakreif: I am always worried from this, in case 'read' will return a single stream if its overloaded somehow: 
```
>>> a=b'12'
>>> list(a)
[49, 50]
```

barakreif: I have seen the above happen when 2to3 conversion was mishandled for bytestreams.

asottile: yeah, fortunately here the `__iter__` function from `IO[bytes]` will give bytes (and mypy knows this)

asottile:
One tiny thing, I edited your description to say "Resolves #..." -- that way when the PR is merged it automatically closes the issue that it is referencing

Thanks again for the patch!  It's greatly appreciated :)

asottile:

PR #421, opened 2019-10-25
trailing-whitespace: add option for custom chars to strip

iconmaster5326:
This PR adds a new flag to the `trailing-whitespace` hook, `--chars`, which allows you to specify a custom set of characters to pass to `rstrip`.

This flag was motivated by Common Lisp, in which a common convention is to use `\f` to delineate sections of code. `trailing-whitespace` considers these by default whitespace, and so strips out the section delimiters users want to preserve. In repositories where this convention is used, a custom value for `--chars` can be specified. This hook is general enough to be used in other use cases if need be, however.

### DIFF
--- a/pre_commit_hooks/trailing_whitespace_fixer.py
+++ b/pre_commit_hooks/trailing_whitespace_fixer.py
@@ -94,7 +94,7 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
         if _fix_file(
             filename,
             md,
-            None if args.chars is None else bytes(args.chars.encode('utf-8')),
+            None if args.chars is None else args.chars.encode('utf-8'),
         ):
             print('Fixing {}'.format(filename))
             return_code = 1
--- a/pre_commit_hooks/trailing_whitespace_fixer.py
+++ b/pre_commit_hooks/trailing_whitespace_fixer.py
@@ -29,13 +29,15 @@ def _process_line(line, is_markdown, chars_to_strip):
     # type: (bytes, bool, Optional[bytes]) -> bytes
     if line[-2:] == b'\r\n':
         eol = b'\r\n'
+        line = line[:-2]
     elif line[-1:] == b'\n':
         eol = b'\n'
+        line = line[:-1]
     else:
         eol = b''
     # preserve trailing two-space for non-blank lines in markdown files
-    if is_markdown and (not line.isspace()) and line.endswith(b'  ' + eol):
-        return line.rstrip(chars_to_strip) + b'  ' + eol
+    if is_markdown and (not line.isspace()) and line.endswith(b'  '):
+        return line[:-2].rstrip(chars_to_strip) + b'  ' + eol
     return line.rstrip(chars_to_strip) + eol
 
 
--- a/pre_commit_hooks/trailing_whitespace_fixer.py
+++ b/pre_commit_hooks/trailing_whitespace_fixer.py
@@ -7,10 +7,15 @@ from typing import Optional
 from typing import Sequence
 
 
-def _fix_file(filename, is_markdown):  # type: (str, bool) -> bool
+def _fix_file(filename, is_markdown, chars_to_strip):
+    # type: (str, bool, Optional[bytes]) -> bool
     with open(filename, mode='rb') as file_processed:
         lines = file_processed.readlines()
-    newlines = [_process_line(line, is_markdown) for line in lines]
+    newlines = [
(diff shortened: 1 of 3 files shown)


pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -7,10 +7,15 @@
 from typing import Sequence
 
 
-def _fix_file(filename, is_markdown):  # type: (str, bool) -> bool
+def _fix_file(filename, is_markdown, chars_to_strip):
```
asottile: I would maybe just call this argument `chars` to match the commandline option -- you'll need to reflow less too :+1:

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -50,6 +58,11 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
             'default: %(default)s'
         ),
     )
+    parser.add_argument(
+        '--chars',
+        help='The set of characters to strip from the end of lines.  '
+        'Defaults to all whitespace characters.',
```
asottile: I would write this as:

```
help=(
    '...'
    '...'
),
```

just so it's slightly easier to read with all the text lined up

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -78,7 +91,11 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
     for filename in args.filenames:
         _, extension = os.path.splitext(filename.lower())
         md = all_markdown or extension in md_exts
-        if _fix_file(filename, md):
+        if _fix_file(
```
asottile: this line is doing a bit much, maybe instead:

```python
chars = args.chars.encode('UTF-8') if args.chars else None

...

    ...

    if _fix_file(filename, md, chars):
        ...
```
```

(this also lifts the argument parsing out of the loop instead of doing it N times)

File: pre_commit_hooks/trailing_whitespace_fixer.py [hunk 4]
```diff
@@ -20,17 +25,20 @@ def _fix_file(filename, is_markdown):  # type: (str, bool) -> bool
...
+        line = line[:-1]
     else:
         eol = b''
     # preserve trailing two-space for non-blank lines in markdown files
-    if is_markdown and (not line.isspace()) and line.endswith(b'  ' + eol):
-        return line.rstrip() + b'  ' + eol
-    return line.rstrip() + eol
+    if is_markdown and (not line.isspace()) and line.endswith(b'  '):
+        return line[:-2].rstrip(chars_to_strip) + b'  ' + eol
+    return line.rstrip(chars_to_strip) + eol
```
asottile: the change in the markdown logic isn't tested -- can you demo that with a test?

pre_commit_hooks/trailing_whitespace_fixer.py
```diff
@@ -78,7 +91,11 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
     for filename in args.filenames:
         _, extension = os.path.splitext(filename.lower())
         md = all_markdown or extension in md_exts
-        if _fix_file(filename, md):
+        if _fix_file(
+            filename,
+            md,
+            None if args.chars is None else bytes(args.chars.encode('utf-8')),
```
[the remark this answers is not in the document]
iconmaster5326: Good catch.

```suggestion
            None if args.chars is None else args.chars.encode('utf-8'),
```

pre_commit_hooks/trailing_whitespace_fixer.py
iconmaster5326: Addressed in commit a33a8f0.

pre_commit_hooks/trailing_whitespace_fixer.py
iconmaster5326: Addressed in commit a33a8f0.

pre_commit_hooks/trailing_whitespace_fixer.py
iconmaster5326: Addressed in commit a33a8f0.

pre_commit_hooks/trailing_whitespace_fixer.py
iconmaster5326: Addressed in commit 0114962.

asottile:

PR #425, opened 2019-11-13
Fixed sorting for non editable requirement urls

vinayinvicible:


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -1,31 +1,41 @@
...
-from typing import IO
+import re
+from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Sequence
 
+from pip._internal.download import PipSession
+from pip._internal.req import InstallRequirement
+from pip._internal.req.req_file import parse_requirements
```
asottile: we can't import private implementation detail from `pip` -- this won't work generically and won't continue to work in the future

vinayinvicible: I do agree with that.
If we don't use this, we'll have to handle each possible variation of requirements file.
IMO using this is better than reinventing the wheel.

asottile: I think it's fine to only support a subset of requirement types (as we're already doing) and add particular patterns as they are encountered

vinayinvicible: I've made the required changes.

pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -21,8 +21,9 @@ def __init__(self):  # type: () -> None
     @property
     def name(self):  # type: () -> bytes
         assert self.value is not None, self.value
-        if self.value.startswith(b'-e '):
-            return self.value.lower().partition(b'=')[-1]
+        for egg in (b'#egg=', b'&egg='):
+            if egg in self.value:
+                return self.value.lower().partition(egg)[-1]
```
asottile: it seems that `egg=` is optional when non-editable:

```
git+https://github.com/asottile/astpretty
```

```console
$ pip install -r t.txt
...
Successfully installed astpretty-1.6.0
```

vinayinvicible: yes, egg is optional when installing in non-editable mode.
But in that case we don't really have a way determine the package name.

PR #429, opened 2019-12-04
fix regular expression for test files

rrauenza:
Noticed this when reading the code -- the regular expression doesn't escape the . in the .py


asottile:
thanks for the fix! mind adding a quick test? should be easy to show with like `foo_py`

rrauenza:
Ok, submitted one.  I cheated for the first fix and just did it in github's online editor :)

PR #431, opened 2019-12-19
Add `--check` support to EOF and Trailing Whitespace fixers

ngie-eign:
This change adds an advisory mode via `--check` that only warns of
formatting issues with files, but does not address them.

This support is desirable because--while I don't mind the automagic
changes when done in a mechanical way--some individuals who I
described the current behavior of these fixers to were a bit uneasy
about the magic that went along with them. Adding `--check` so
others can opt out (similar to `black --check`) is a compromise on
this front.

Signed-off-by: Enji Cooper <yaneurabeya@gmail.com>

### DIFF
--- a/pre_commit_hooks/end_of_file_fixer.py
+++ b/pre_commit_hooks/end_of_file_fixer.py
@@ -9,7 +9,7 @@ from typing import Optional
 from typing import Sequence
 
 
-def fix_file(file_obj):  # type: (IO[bytes]) -> int
+def fix_file(file_obj, apply_fixes=True):  # type: (IO[bytes], bool) -> int
     # Test for newline at end of file
     # Empty files will throw IOError here
     try:
@@ -20,8 +20,9 @@ def fix_file(file_obj):  # type: (IO[bytes]) -> int
     # last_character will be '' for an empty file
     if last_character not in {b'\n', b'\r'} and last_character != b'':
         # Needs this seek for windows, otherwise IOError
-        file_obj.seek(0, os.SEEK_END)
-        file_obj.write(b'\n')
+        if apply_fixes:
+            file_obj.seek(0, os.SEEK_END)
+            file_obj.write(b'\n')
         return 1
 
     while last_character in {b'\n', b'\r'}:
@@ -29,8 +30,9 @@ def fix_file(file_obj):  # type: (IO[bytes]) -> int
         if file_obj.tell() == 1:
             # If we've reached the beginning of the file and it is all
             # linebreaks then we can make this file empty
-            file_obj.seek(0)
-            file_obj.truncate()
+            if apply_fixes:
+                file_obj.seek(0)
+                file_obj.truncate()
             return 1
 
         # Go back two bytes and read a character
@@ -45,8 +47,9 @@ def fix_file(file_obj):  # type: (IO[bytes]) -> int
         if remaining == sequence:
             return 0
         elif remaining.startswith(sequence):
-            file_obj.seek(position + len(sequence))
-            file_obj.truncate()
+            if apply_fixes:
+                file_obj.seek(position + len(sequence))
+                file_obj.truncate()
             return 1
 
     return 0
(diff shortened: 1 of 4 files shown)


asottile:
while I appreciate the effort here, this really isn't in the spirit of these two hooks and not something I'd like to support / encourage

I'd encourage you to either:

1. use a `pygrep` hook to enforce this instead
2. encourage your teammates of the safety of these hooks

these two are basically the [most universally used hooks](https://github.com/chriskuehl/pre-commit-stats) and are very battle tested

the ergonomics of a checker is *way worse* than that of a fixer, to the point where people will get frustrated with the framework

ngie-eign:
(Replying out-of-order to the previous statement)

> these two are basically the most universally used hooks and are very battle tested

While I agree based on the data you presented, but data is entirely opt-in, so there's a good chance that the information presented isn't complete.

> the ergonomics of a checker is way worse than that of a fixer, to the point where people will get frustrated with the framework

> while I appreciate the effort here, this really isn't in the spirit of these two hooks and not something > I'd like to support / encourage
>
> I'd encourage you to either:
>
>   1. use a pygrep hook to enforce this instead
>   2. encourage your teammates of the safety of these hooks

I got a lot of pushback on 2. My goal is to get things adopted (right now we have nothing) in a very large codebase and in order to sell these tools to my organization, I need a way to do a phased rollout. Running the linters through a check mode enables me to do that.

I would mind a bit less if there was something in the fixers that built off of pygrep (or alternatively hooked into pygrep), but that doesn't seem to be the case today.

asottile:
they're very easy to vet against the entire codebase already -- run `pre-commit run --all-files` and use `git diff` to see what changes

asottile:
(by the way, the data above was scraped from github search a few years ago -- while it's slightly out of date now and it's entirely possible that closed-source shops are using pre-commit in a slightly different manner -- I'm fairly certain that the open source usage is quite representative of most usage)

asottile:
an easier route for your organization might be to utilize your fork of these hooks until your organization is comfortable -- since `pre-commit` uses `git` as the distribution mechanism this is as easy as pointing at your repository instead

ngie-eign:
> they're very easy to vet against the entire codebase already -- run `pre-commit run --all-files` and use git diff to see what changes.

I'd rather not do that, especially since this project is effectively a fork of another upstream project where stylization doesn't necessarily have the same degree of sense:
```
$ find . -type f | wc -l
315290
```
> an easier route for your organization might be to utilize your fork of these hooks until your organization is comfortable -- since `pre-commit` uses `git` as the distribution mechanism this is as easy as pointing at your repository instead
Right now `pre-commit` only allows labels, not branches, in the config. In order to avoid hacking things too much internally, deviating from an upstream, I'd rather have these changes sitting out on a branch, e.g.,
```
* upstream/master ----> internal-fork/add-check-to-eof-and-ws-fixers
```
Also, the point was to involve less magic and leverage opensource more effectively. It's not my job to maintain this tool indefinitely, and the amount of magic/knowledge I have to transfer on to someone else makes the likelihood of adoption lower.

ngie-eign:
> Right now `pre-commit` only allows labels, not branches, in the config. In order to avoid hacking things too much internally, deviating from an upstream, I'd rather have these changes sitting out on a branch, e.g.,

Reminder to self: `git tag` works via commit objects, not branches, so this is less of an issue. However, my concern about maintaining a fork indefinitely seems like a non-starter.

I'll think about this some more and come back to it in another week.

asottile:
> I'd rather not do that, especially since this project is effectively a fork of another upstream project where stylization doesn't necessarily have the same degree of sense:

another way to ratchet up linting is to select a small subset of your codebase to vet it on -- this was actually made much easier in [pre-commit v1.21.0](https://github.com/pre-commit/pre-commit/releases/tag/v1.21.0) which adds a top level [`files`](https://pre-commit.com/#pre-commit-configyaml---top-level) to allow selection of a particular set of files.  (this was previously available on the hook level but the top level setting reduces the amount of copy paste needed to achieve that)

> Also, the point was to involve less magic and leverage opensource more effectively. It's not my job to maintain this tool indefinitely

hehehe -- that's my primary reason for not accepting this as-is, I don't really want to support a use case that I (1) don't really agree with and (2) most people won't use

> Reminder to self: `git tag` works via commit objects,

ah yes, at $different_employer we used `v1.x.x.post1` for internal forked things which worked reasonably well -- if there were features we needed we'd periodically rebase (and wrote tests, etc.) so it wasn't so bad

PR #433, opened 2019-12-31
run-sub-hooks hook added

rooterkyberian:
Looking for comments right now, but ideally if it works out I would love to see it merged;)

Whole thing was inspired by:
https://github.com/pre-commit/pre-commit/issues/731#issuecomment-376945745

It can somewhat solve two problems:
* running sub pre-commits with their own configurations
* running sub pre-commits limited to a subdirectory
  * addresses monorepo https://github.com/pre-commit/pre-commit/issues/466
  * allows for organizing hooks, although the cost of this hook is probably too high to make this a good use case

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -192,3 +192,9 @@
     language: python
     types: [text]
     stages: [commit, push, manual]
+-   id: run-sub-hooks
+    name: Run Subdirectory Hooks
+    description: Runs subprocess with alternative pre-commit configuration.
+    entry: run-sub-hooks
+    language: python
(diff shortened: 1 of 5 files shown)


asottile:
sorry, I don't want to support this directly as it is a worse user experience and violates some of the cacheability guarantees

thanks for the patch nonetheless

rooterkyberian:
can you to point me to what cache weirdness can occur?

asottile:
for example, [this won't work](https://github.com/pre-commit/action/blame/14f08406dfbbe34836ec032fb0f10c1004fb0091/README.md#L29-L32)

PR #442, opened 2020-02-03
Allow linking directly to each hook

kimsey0:
I don't know if this is something that you want to maintain, so feel free to reject the pull request. I just wanted to suggest one of the hooks from here to my team, and wished I could like directly to the correct one like https://github.com/pre-commit/pre-commit-hooks#mixed-line-ending. This pull request allows that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,99 +22,153 @@ Add this to your `.pre-commit-config.yaml`
 
 ### Hooks available
 
-- <a name="check-added-large-files">`check-added-large-files`</a> - Prevent giant files from being committed.
-    - Specify what is "too large" with `args: ['--maxkb=123']` (default=500kB).
-    - If `git-lfs` is installed, lfs files will be skipped
-      (requires `git-lfs>=2.2.1`)
-- <a name="check-ast">`check-ast`</a> - Simply check whether files parse as valid python.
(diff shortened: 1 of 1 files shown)


asottile:
I like the idea, but is there a way to do this with vanilla markdown?

kimsey0:
Not that I can find, unfortunately. The [GitHub Flavored Markdown Spec](https://github.github.com/gfm/) doesn't seem to include any way to define named anchors, and none of [the syntaxes I've seen suggested on the web](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Ca+name%3D%22anchorOne%22%3E%3C%2Fa%3Eone%0A%0A%5B%5D()%7B%3Aname%3D%27anchorTwo%27%7Dtwo%0A%0A%7B%3Aname%3D%27anchorThree%27%7Dthree%0A%0A%7B%23anchorFour%7Dfour) work.

asottile:
hmmm a good compromise would be to convert them to headings I think, let's do that (and use some small heading like
```markdown
#### `hook-name`
```

kimsey0:
I converted them to headings. It does introduce a newline after each hook name, which makes the rendered page longer, but hopefully that's OK.

I couldn't quite figure out what the max line length was supposed to be, but I removed some line breaks that seemed out of place after splitting up each hook description into a heading and a body.

asottile:
thanks!

kimsey0:
Thank you!

PR #446, opened 2020-02-05
Handle complex md exts

### LINKED ISSUE #445: trailing-whitespace extension doesn't allow for md.j2
```
error: bad --markdown-linebreak-ext extension '.md.j2' (has . / \ :)
  (probably filename; use '--markdown-linebreak-ext=EXT')```

### PR DESCRIPTION
**Pull request by Natim**
Fixes #445

### DIFF
--- a/pre_commit_hooks/trailing_whitespace_fixer.py
+++ b/pre_commit_hooks/trailing_whitespace_fixer.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 
 import argparse
-import os
 import sys
 from typing import Optional
 from typing import Sequence
@@ -78,17 +77,17 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
 
     # reject probable "eaten" filename as extension: skip leading '.' with [1:]
     for ext in md_exts:
-        if any(c in ext[1:] for c in r'./\:'):
+        if any(c in ext[1:] for c in r'/\:'):
             parser.error(
-                'bad --markdown-linebreak-ext extension {!r} (has . / \\ :)\n'
+                'bad --markdown-linebreak-ext extension {!r} (has / \\ :)\n'
                 "  (probably filename; use '--markdown-linebreak-ext=EXT')"
                 .format(ext),
             )
     chars = None if args.chars is None else args.chars.encode('utf-8')
     return_code = 0
     for filename in args.filenames:
-        _, extension = os.path.splitext(filename.lower())
-        md = all_markdown or extension in md_exts
+        in_md_exts = any([filename for ext in md_exts if filename.endswith(ext)])
+        md = all_markdown or in_md_exts
         if _fix_file(filename, md, chars):
             print('Fixing {}'.format(filename))
             return_code = 1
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pre_commit_hooks/trailing_whitespace_fixer.py [hunk 1]
```diff
@@ -78,17 +77,17 @@ def main(argv=None):  # type: (Optional[Sequence[str]]) -> int
...
+                'bad --markdown-linebreak-ext extension {!r} (has / \\ :)\n'
                 "  (probably filename; use '--markdown-linebreak-ext=EXT')"
                 .format(ext),
             )
     chars = None if args.chars is None else args.chars.encode('utf-8')
     return_code = 0
     for filename in args.filenames:
-        _, extension = os.path.splitext(filename.lower())
-        md = all_markdown or extension in md_exts
+        in_md_exts = any([filename for ext in md_exts if filename.endswith(ext)])
```
asottile: this breaks for the reason I listed here: https://github.com/pre-commit/pre-commit-hooks/issues/445#issuecomment-582520693

Natim: I don't think so, please add a test to show this behaviour.

asottile: yes, please do so

Natim: :+1:

asottile:
I see, this was guarded by `lstrip('.')` before

one thing this does that breaks is this is now legal (and a pretty easy footgun):

### after

```console
$ trailing-whitespace-fixer --markdown setup.py
$
```

### before
```console
$ trailing-whitespace-fixer --markdown setup.py
usage: trailing-whitespace-fixer [-h]
                                 [--markdown-linebreak-ext *|EXT[,EXT,...]]
                                 [--chars CHARS]
                                 [filenames [filenames ...]]
trailing-whitespace-fixer: error: bad --markdown-linebreak-ext extension '.setup.py' (has . / \ :)
  (probably filename; use '--markdown-linebreak-ext=EXT')
$
```

Natim:
Yeah but that's what the user asked for.


We can also enforce extensions to start with a `.` to mitigate the issue.

Or add a warning if there is not `.` at the beginning.

Le mer. 5 févr. 2020 à 20:15, Anthony Sottile <notifications@github.com> a
écrit :

asottile:
that's backwards incompatible, I don't think this patch is possible to do in a backward compatible way, warnings aren't helpful as they are entirely hidden

Natim:
Let's do a breaking change then.

People using pre-commit tag the version of their hooks already so it
wouldn't be to much of a itch it guess.

Le mer. 5 févr. 2020 à 20:25, Anthony Sottile <notifications@github.com> a
écrit :

asottile:
I'd rather not

Natim:
If you are really against it feel free to close the PR. I am ok with your
arguments and can use a fork for my need.

Thank you for your kind review and consideration.

Le mer. 5 févr. 2020 à 20:31, Anthony Sottile <notifications@github.com> a
écrit :

asottile:
makes sense, thanks for this patch but I don't think we can take this, fortunately pre-commit makes it easy to use a fork

Natim:
👍

Le mer. 5 févr. 2020 à 21:50, Anthony Sottile <notifications@github.com> a
écrit :

Natim:
What if I keep the current behavior with dots in extensions and I add a new option to allow them `--allow-dots-in-md-ext`

It would handle the breaking change issue. V what do you think?

asottile:
does this work?

```yaml
    hooks:
    -   id: trailing-whitespace
        exclude: \.md\.j2$
    -   id: trailing-whitespace
        files: \.md\.j2$
        args: [--markdown-linebreak-ext=*]
```

PR #448, opened 2020-02-06
Add check for unicode replacement characters

scop:
Often a sign of charset confusion/corruption.


asottile:
This probably is possible / simpler / makes more in https://github.com/pre-commit/pygrep-hooks

scop:
Sure, why not -> https://github.com/pre-commit/pygrep-hooks/pull/26

asottile:
awesome, we'll continue in pygrep-hooks :)

Issue #449, opened 2020-02-12
detect-aws-credentials false positives for empty AWS variables


begoon:
When there are AWS credentials variables with empty values are in the environment, for example:
```
AWS_SECRET_ACCESS_KEY=
AWS_ACCESS_KEY_ID=
```

`detect_aws_credentials` defects a false positive issue, for example:

```
Detect AWS Credentials...................................................Failed
hookid: detect-aws-credentials

AWS secret found in some/file/admin.py: ****************************
```

It happens because the `key` value is an empty string "", and obviously the empty values "" can be "detected" in every file.

The following fix allows to skip searching for empty values of the keys.
```
index fe18f4d..aae0734 100644
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -84,7 +84,7 @@ def check_file_for_aws_keys(
             for key in keys:
                 # naively match the entire file, low chance of incorrect
                 # collision
-                if key in text_body:
+                if key and key in text_body:
                     bad_files.append(BadFile(filename, key[:4].ljust(28, '*')))
     return bad_files
```

begoon:
Version 2.5.0.

WhyNotHugo:
AFAIK, all valid aws keys start with `AKIA`, so that might also be a more reliable hint to detect them.

asottile:
this was already resolved in #450

PR #450, opened 2020-02-12
Skip empty key variables in detect_aws_credentials

begoon:
This PR resolves the following - issue https://github.com/pre-commit/pre-commit-hooks/issues/449

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -31,7 +31,7 @@ def get_aws_secrets_from_env() -> Set[str]:
     for env_var in (
         'AWS_SECRET_ACCESS_KEY', 'AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN',
     ):
-        if env_var in os.environ and os.environ[env_var]:
+        if os.environ.get(env_var):
             keys.add(os.environ[env_var])
     return keys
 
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -84,7 +84,7 @@ def check_file_for_aws_keys(
             for key in keys:
                 # naively match the entire file, low chance of incorrect
                 # collision
-                if key in text_body:
+                if key and key in text_body:
```
asottile: can you add a quick test for this to demonstrate the change -- definitely agree with this fix though :+1:

begoon: I moved the change to another place, and also added a test.

pre_commit_hooks/detect_aws_credentials.py
```diff
@@ -31,7 +31,7 @@ def get_aws_secrets_from_env() -> Set[str]:
     for env_var in (
         'AWS_SECRET_ACCESS_KEY', 'AWS_SECURITY_TOKEN', 'AWS_SESSION_TOKEN',
     ):
-        if env_var in os.environ:
+        if env_var in os.environ and os.environ[env_var]:
```
asottile: ~slightly simpler: `if os.environ.get(env_var)`

begoon: Fixed

asottile:
thanks!

PR #479, opened 2020-05-17
Support checking unicode TOML

### LINKED ISSUE #474: Bug check-toml ASCII on Windows
I run pre-commit-hooks on my CI using Linux, Windows and Mac. When I have ASCII characters on it in Windows, it gives me an error:

`UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 148: character maps to <undefined>`

The ASCII character is in the author name from this file: https://github.com/staticdev/toml-validator/blob/master/pyproject.toml

When I remove the Á from the name, the error vanishes. This is a bug.

mxr:
Resolves #474

### DIFF
--- a/pre_commit_hooks/check_toml.py
+++ b/pre_commit_hooks/check_toml.py
@@ -13,8 +13,7 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
     retval = 0
     for filename in args.filenames:
         try:
-            with open(filename) as f:
-                toml.load(f)
+            toml.load(filename)
         except toml.TomlDecodeError as exc:
             print(f'{filename}: {exc}')
             retval = 1
--- a/tests/check_toml_test.py
+++ b/tests/check_toml_test.py
@@ -1,7 +1,7 @@
 from pre_commit_hooks.check_toml import main
 
 
-def test_toml_good(tmpdir):
+def test_toml_bad(tmpdir):
     filename = tmpdir.join('f')
     filename.write("""
 key = # INVALID
@@ -12,7 +12,7 @@ key = # INVALID
     assert ret == 1
 
 
-def test_toml_bad(tmpdir):
+def test_toml_good(tmpdir):
     filename = tmpdir.join('f')
     filename.write(
         """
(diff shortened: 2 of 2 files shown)


tests/check_toml_test.py
```diff
@@ -23,6 +23,7 @@ def test_toml_bad(tmpdir):
 [owner]
 name = "John"
 dob = 1979-05-27T07:32:00-08:00 # First class dates
+favorite_letter = "Á"
```
asottile, line 26: this test will fail in the same way as the issues is -- I'd suggest writing a separate test which does `.write_binary('...'.encode('UTF-8')`

note that `Á` is not a good choice for a UTF-8-compatible test since that's encodable in latin1 / cp1252 I usually pick snowman (`\u2603`) or upside down face (`\U0001f643`)

mxr: Sure, will hack on this

pre_commit_hooks/check_toml.py
```diff
@@ -13,7 +13,7 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
     retval = 0
     for filename in args.filenames:
         try:
-            with open(filename) as f:
+            with open(filename, encoding='UTF-8') as f:
                 toml.load(f)
```
mxr, line 17: Was looking at the `toml.load` implementaion and this could just be `tom.load(filename)`?

asottile: if it works go for it

mxr: Cool, it means we don't have to manually specify the encoding

staticdev:
@mxr something strange.. I pointed my repo to v3.0.1 with your fix and I am getting the same error: https://github.com/staticdev/toml-validator/pull/30

asottile:
@staticdev I see
```
Check Toml...............................................................Passed
```

you have a failure in `prettier` though 😉

staticdev:
Wow.. I am sorry! I just realized that.

PR #480, opened 2020-05-17
Check git mode on Windows

### LINKED ISSUE #435: check-executables-have-shebangs not correct on windows
does not seem to work right on windows

i am dealing with a primarily linux / python project but on a windows machine.

i have set `git config core.filemode false`

i created a new file and stage and i verify that filemode is 644:
```
>git ls-files -s newfile.py
100644 3edd36f71bf2081c70a0eaf39dec6980d0a9f791 0       newfile.py
```

but hook still fails

```
hookid: check-executables-have-shebangs

newfile.py: marked executable but has no (or invalid) shebang!
  If it isn't supposed to be executable, try: chmod -x newfile.py
  If it is supposed to be executable, double-check its shebang.
```

why is this file causing error?

mxr:
Resolves #435

### DIFF
--- a/tests/check_executables_have_shebangs_test.py
+++ b/tests/check_executables_have_shebangs_test.py
@@ -1,8 +1,19 @@
+import os
+import sys
+
 import pytest
 
+from pre_commit_hooks import check_executables_have_shebangs
 from pre_commit_hooks.check_executables_have_shebangs import main
+from pre_commit_hooks.util import cmd_output
+
+skip_win32 = pytest.mark.skipif(
+    sys.platform == 'win32',
+    reason="non-git checks aren't relevant on windows",
+)
 
 
+@skip_win32  # pragma: win32 no cover
 @pytest.mark.parametrize(
     'content', (
         b'#!/bin/bash\nhello world\n',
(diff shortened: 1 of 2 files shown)


mxr:
@asottile I  can't request a review in the GitHub UI but this should be ready for review now

mxr:
Awesome!

tests/check_executables_have_shebangs_test.py
```diff
@@ -33,3 +44,67 @@ def test_bad_shebang(content, tmpdir, capsys):
...
     assert stderr.startswith(f'{path}: marked executable but')
+
+
+def test_check_git_filemode_passing(tmpdir):
+    with tmpdir.as_cwd():
+        cmd_output('git', 'init', '.')
+
+        f = tmpdir.join('f')
+        f.write('#!/usr/bin/env bash')
+        f_path = str(f)
```
mxr, line 55: The other tests use `.strpath`, is one preferred over the other?

asottile: apparently `strpath` is not documented despite being a public attribute so I've mostly switched to `str(...)` now that I don't care about python2

mxr: TIL

PR #482, opened 2020-05-17
pytype: pass str filename to xml.sax.parse, not BinaryIO

zsxoff:
Perhaps this will avoid `FileNotFoundError`.


asottile:
what does this fix? can you demonstrate with a test?

asottile:
`xml.sax.parse(str, ...)` has potential security problems, notably it will download anything that looks like a url

PR #483, opened 2020-05-18
Parse more operators in requirements

### LINKED ISSUE #331: requirements-txt-fixer does not parse all requirements
While working on #330 I noticed that the requirements-txt-fixer hook only parses a small subset of requirement specifiers. Take this example:

    pyramid-foo==1
    pyramid>=2

`pyramid` should sort before `pyramid-foo`, but the hook interprets the `>=2` as part of the name.

According to [PEP 508](https://www.python.org/dev/peps/pep-0508/) there are a couple of things that can be specified after the name (extras, versions, markers) so maybe an approach would be to use a regular expression to extract only the package name?

mxr:
Resolves #331

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -1,4 +1,5 @@
 import argparse
+import re
 from typing import IO
 from typing import List
 from typing import Optional
@@ -10,6 +11,9 @@ FAIL = 1
 
 
 class Requirement:
+    UNTIL_COMPARISON = re.compile(b'={2,3}|!=|~=|>=?|<=?')
+    UNTIL_SEP = re.compile(rb'[^;\s]+')
+
     def __init__(self) -> None:
         self.value: Optional[bytes] = None
         self.comments: List[bytes] = []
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -10,18 +11,31 @@
 
 
 class Requirement:
+    UNTIL_COMPARISON = re.compile(b'={2,3}|!=|~=|>=?|<=?')
```
asottile, line 14: is this supposed to match `===` ?

pre_commit_hooks/requirements_txt_fixer.py
```diff
@@ -10,18 +11,31 @@
...
+    UNTIL_SEP = re.compile(rb'\S[^;\s]+')
```
asottile, line 15: is this different from `[^;\s]+`?

pre_commit_hooks/requirements_txt_fixer.py
mxr: Yes, we match it here too https://github.com/asottile/setup-cfg-fmt/blob/c3a291b4615a1ed07b275dd37d20df9ca02f8880/setup_cfg_fmt.py#L228

pre_commit_hooks/requirements_txt_fixer.py
mxr: Good call, I got confused parsing the `--hash` lines

pre_commit_hooks/requirements_txt_fixer.py
asottile: huh, TIL I guess

PR #489, opened 2020-05-20
Support attribute docstrings

### LINKED ISSUE #159: check-docstring-first fails positive on attribute docstrings
Attribute docstrings, as [defined in PEP-257](https://www.python.org/dev/peps/pep-0257/#id15) are strings immediately following top level module names.  For example:

```
name = "my variable"
"""the name used by the module for x, y, and z"""
```

The check-docstring-first incorrectly identifies this attribute docstring as a module docstring, indicating either `Multiple module docstrings (first docstring on line N).` or `Module docstring appears after code (code seen on line N).`.

In light of this issue, I'm not sure the `check-docstring-first` check has much value at all. At the very least, the check should not be failing when a docstring follows an assignment.

mxr:
Resolves #159

### DIFF
--- a/pre_commit_hooks/check_docstring_first.py
+++ b/pre_commit_hooks/check_docstring_first.py
@@ -11,7 +11,7 @@ NON_CODE_TOKENS = frozenset((
 ))
 
 
-def check_docstring_first(src: bytes, filename: str = '<unknown>') -> int:
+def check_docstring_first(src: bytes, filename: str) -> int:
     """Returns nonzero if the source has what looks like a docstring that is
     not at the beginning of the source.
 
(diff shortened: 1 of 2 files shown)


asottile:
thanks for the patch!

I'm a little torn on this since I think it's ~slightly better to use comments than strings: https://github.com/pre-commit/pre-commit-hooks/issues/159#issuecomment-559886109 (and I'm pretty sure they work?)

mxr:
Happy to close it if there's a workaround

asottile:
> 
> 
> Happy to close it if there's a workaround

I would need to confirm that `#:` works -- I don't have a sphinx project of my own to test it on though 🤔

mxr:
@asottile should we close this?

mbukatov:
Using a doc string instead of `#:` comment also allows to use rst syntax or explain the variable in multiple lines.

asottile:
yeah I think I'll just close this

PR #492, opened 2020-06-01
Add hook to reject verbose/talkative curl commands in files

debuiteb:


asottile:
hello, thanks for the contribution but I don't think this is generic / applicable enough to be in pre-commit/pre-commit-hooks

also you might find the [`pygrep`](https://pre-commit.com/#pygrep) language more convenient for implementing simple regex-based hooks

thanks again for the contribution -- you may also consider making a separate repository where you can house this hook and other people can reuse your work there!

PR #493, opened 2020-06-04
Add --no-fail arg to trailing whitespace

adeebshihadeh:


asottile:
could you fill out the description or explain why this is useful?

PR #494, opened 2020-06-08
Add check_path hook

Bysmyyr:
Adds a hook which can be used for preventing commits to paths specified
by regex.


asottile:
there's no need for this as [`language: fail`](https://pre-commit.com/#fail) does exactly this

PR #503, opened 2020-07-09
Fix reference to trailing whitespace fixer

ondrejzacha:
The command-line tool is named trailing-whitespace-fixer, not trailing-whitespace.
Adjusted the README to account for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ very specific format.  You must opt in to this by setting [`files`](https://pre-
 ```
 
 
-#### `trailing-whitespace`
+#### `trailing-whitespace-fixer`
 Trims trailing whitespace.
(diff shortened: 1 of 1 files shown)


asottile:
the hook is named `trailing-whitespace` so the README is correct and this patch is incorrect (as you can see from the test failure)

ondrejzacha:
You're right, I missed the yaml file. However, the console script for `trailing-whitespace` is called `trailing-whitespace-fixer`.
I compared `.pre-commit-hooks.yaml` and `setup.cfg` and next to this, also the console script for `debug-statements` is called `debug-statement-hook`. Other hooks have matching console script names. Is this intended?

asottile:
it's too late to really fix them since some depend on both names.  it's better from a sanity perspective that they match but not necesssary

PR #511, opened 2020-08-02
New hook 'destroyed-symlinks'

m-khvoinitsky:
The hook detects symlinks which are changed to regular files with a content of a path which that symlink was pointing to.
This usually happens on Windows in case when user without a permission for creating symlinks clones repository with symlinks.

~~Some implementation notes: `bytes` is used for filenames in order to avoid having any troubles with encodings. After all, the hook check for destroyed symlinks, not for bad file encoding. `;)`~~

P. S. If you don't think that this hook would be useful for general audience, I'm fine with hosting it in the separate repo. Just wanted to try.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -100,6 +100,12 @@
     entry: debug-statement-hook
     language: python
     types: [python]
+-   id: destroyed-symlinks
+    name: Detect Destroyed Symlinks
+    description: Detects symlinks which are changed to regular files with a content of a path which that symlink was pointing to.
+    entry: destroyed-symlinks
+    language: python
+    types: [file]
 -   id: detect-aws-credentials
     name: Detect AWS Credentials
     description: Detects *your* aws credentials from the aws cli credentials file
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ Attempts to load all yaml files to verify syntax.
 #### `debug-statements`
 Check for debugger imports and py37+ `breakpoint()` calls in python source.
 
+#### `destroyed-symlinks`
+Detects symlinks which are changed to regular files with a content of a path
+which that symlink was pointing to.
+This usually happens on Windows when a user clones a repository that has
+symlinks but they do not have the permission to create symlinks.
+
 #### `detect-aws-credentials`
 Checks for the existence of AWS secrets that you have set up with the AWS CLI.
 The following arguments are available:
--- a/pre_commit_hooks/destroyed_symlinks.py
+++ b/pre_commit_hooks/destroyed_symlinks.py
@@ -1,0 +1,96 @@
+import argparse
+import shlex
+import subprocess
+from typing import List
+from typing import Optional
+from typing import Sequence
+
+from pre_commit_hooks.util import cmd_output
+from pre_commit_hooks.util import zsplit
+
+ORDINARY_CHANGED_ENTRIES_MARKER = '1'
+PERMS_LINK = '120000'
+PERMS_NONEXIST = '000000'
+
+
+def find_destroyed_symlinks(files: Sequence[str]) -> List[str]:
+    destroyed_links: List[str] = []
+    if not files:
+        return destroyed_links
+    for line in zsplit(
+        cmd_output('git', 'status', '--porcelain=v2', '-z', '--', *files),
+    ):
+        splitted = line.split(' ')
+        if splitted and splitted[0] == ORDINARY_CHANGED_ENTRIES_MARKER:
+            # https://git-scm.com/docs/git-status#_changed_tracked_entries
+            (
+                _, _, _,
+                mode_HEAD,
+                mode_index,
+                _,
+                hash_HEAD,
+                hash_index,
+                *path_splitted,
+            ) = splitted
+            path = ' '.join(path_splitted)
+            if (
+                    mode_HEAD == PERMS_LINK and
+                    mode_index != PERMS_LINK and
+                    mode_index != PERMS_NONEXIST
+            ):
+                if hash_HEAD == hash_index:
+                    # if old and new hashes are equal, it's not needed to check
+                    # anything more, we've found a destroyed symlink for sure
+                    destroyed_links.append(path)
+                else:
+                    # if old and new hashes are *not* equal, it doesn't mean
+                    # that everything is OK - new file may be altered
+                    # by something like trailing-whitespace and/or
+                    # mixed-line-ending hooks so we need to go deeper
+                    SIZE_CMD = ('git', 'cat-file', '-s')
+                    size_index = int(cmd_output(*SIZE_CMD, hash_index).strip())
+                    size_HEAD = int(cmd_output(*SIZE_CMD, hash_HEAD).strip())
+
+                    # in the worst case new file may have CRLF added
+                    # so check content only if new file is bigger
+                    # not more than 2 bytes compared to the old one
+                    if size_index <= size_HEAD + 2:
+                        head_content = subprocess.check_output(
+                            ('git', 'cat-file', '-p', hash_HEAD),
+                        ).rstrip()
+                        index_content = subprocess.check_output(
+                            ('git', 'cat-file', '-p', hash_index),
+                        ).rstrip()
+                        if head_content == index_content:
+                            destroyed_links.append(path)
+    return destroyed_links
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*', help='Filenames to check.')
+    args = parser.parse_args(argv)
+    destroyed_links = find_destroyed_symlinks(files=args.filenames)
+    if destroyed_links:
+        print('Destroyed symlinks:')
+        for destroyed_link in destroyed_links:
+            print(f'- {destroyed_link}')
+        print('You should unstage affected files:')
+        print(
+            '\tgit reset HEAD -- {}'.format(
+                ' '.join(shlex.quote(link) for link in destroyed_links),
+            ),
+        )
+        print(
+            'And retry commit. As a long term solution '
+            'you may try to explicitly tell git that your '
+            'environment does not support symlinks:',
+        )
+        print('\tgit config core.symlinks false')
+        return 1
+    else:
+        return 0
+
+
+if __name__ == '__main__':
+    exit(main())
--- a/pre_commit_hooks/util.py
+++ b/pre_commit_hooks/util.py
@@ -1,5 +1,6 @@
 import subprocess
 from typing import Any
+from typing import List
 from typing import Optional
 from typing import Set
 
@@ -22,3 +23,11 @@ def cmd_output(*cmd: str, retcode: Optional[int] = 0, **kwargs: Any) -> str:
     if retcode is not None and proc.returncode != retcode:
         raise CalledProcessError(cmd, retcode, proc.returncode, stdout, stderr)
     return stdout
+
+
+def zsplit(s: str) -> List[str]:
+    s = s.strip('\0')
+    if s:
+        return s.split('\0')
+    else:
+        return []
--- a/tests/destroyed_symlinks_test.py
+++ b/tests/destroyed_symlinks_test.py
@@ -1,0 +1,74 @@
+import os
+import subprocess
+
+import pytest
+
+from pre_commit_hooks.destroyed_symlinks import find_destroyed_symlinks
+from pre_commit_hooks.destroyed_symlinks import main
+
+TEST_SYMLINK = 'test_symlink'
+TEST_SYMLINK_TARGET = '/doesnt/really/matters'
+TEST_FILE = 'test_file'
+TEST_FILE_RENAMED = f'{TEST_FILE}_renamed'
+
+
+@pytest.fixture
+def repo_with_destroyed_symlink(tmpdir):
+    source_repo = tmpdir.join('src')
+    os.makedirs(source_repo, exist_ok=True)
+    test_repo = tmpdir.join('test')
+    with source_repo.as_cwd():
+        subprocess.check_call(('git', 'init'))
+        os.symlink(TEST_SYMLINK_TARGET, TEST_SYMLINK)
+        with open(TEST_FILE, 'w') as f:
+            print('some random content', file=f)
+        subprocess.check_call(('git', 'add', '.'))
+        subprocess.check_call(
+            ('git', 'commit', '--no-gpg-sign', '-m', 'initial'),
+        )
+        assert b'120000 ' in subprocess.check_output(
+            ('git', 'cat-file', '-p', 'HEAD^{tree}'),
+        )
+    subprocess.check_call(
+        ('git', '-c', 'core.symlinks=false', 'clone', source_repo, test_repo),
+    )
+    with test_repo.as_cwd():
+        subprocess.check_call(
+            ('git', 'config', '--local', 'core.symlinks', 'true'),
+        )
+        subprocess.check_call(('git', 'mv', TEST_FILE, TEST_FILE_RENAMED))
+    assert not os.path.islink(test_repo.join(TEST_SYMLINK))
+    yield test_repo
+
+
+def test_find_destroyed_symlinks(repo_with_destroyed_symlink):
+    with repo_with_destroyed_symlink.as_cwd():
+        assert find_destroyed_symlinks([]) == []
+        assert main([]) == 0
+
+        subprocess.check_call(('git', 'add', TEST_SYMLINK))
+        assert find_destroyed_symlinks([TEST_SYMLINK]) == [TEST_SYMLINK]
+        assert find_destroyed_symlinks([]) == []
+        assert main([]) == 0
+        assert find_destroyed_symlinks([TEST_FILE_RENAMED, TEST_FILE]) == []
+        ALL_STAGED = [TEST_SYMLINK, TEST_FILE_RENAMED]
+        assert find_destroyed_symlinks(ALL_STAGED) == [TEST_SYMLINK]
+        assert main(ALL_STAGED) != 0
+
+        with open(TEST_SYMLINK, 'a') as f:
+            print(file=f)  # add trailing newline
+        subprocess.check_call(['git', 'add', TEST_SYMLINK])
+        assert find_destroyed_symlinks(ALL_STAGED) == [TEST_SYMLINK]
+        assert main(ALL_STAGED) != 0
+
+        with open(TEST_SYMLINK, 'w') as f:
+            print('0' * len(TEST_SYMLINK_TARGET), file=f)
+        subprocess.check_call(('git', 'add', TEST_SYMLINK))
+        assert find_destroyed_symlinks(ALL_STAGED) == []
+        assert main(ALL_STAGED) == 0
+
+        with open(TEST_SYMLINK, 'w') as f:
+            print('0' * (len(TEST_SYMLINK_TARGET) + 3), file=f)
+        subprocess.check_call(('git', 'add', TEST_SYMLINK))
+        assert find_destroyed_symlinks(ALL_STAGED) == []
+        assert main(ALL_STAGED) == 0
--- a/pre_commit_hooks/check_executables_have_shebangs.py
+++ b/pre_commit_hooks/check_executables_have_shebangs.py
@@ -8,16 +8,9 @@ from typing import Sequence
 from typing import Set
 
 from pre_commit_hooks.util import cmd_output
+from pre_commit_hooks.util import zsplit
 
 EXECUTABLE_VALUES = frozenset(('1', '3', '5', '7'))
-
-
-def zsplit(s: str) -> List[str]:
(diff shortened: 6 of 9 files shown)


m-khvoinitsky:
No idea why coverage report is different for pypy, will investigate it.

m-khvoinitsky:
I've finally managed to find a time to dig into pypy issue. The reason was the following. In CPython in code like this:
```python
print('something', file=open(somefile, 'a'))
```
file object losts its last reference it this function call and it is destroyed forcing file buffers to flush. So, for some reason, this doesn't happen in pypy. Workaround would be using explicit `flush=True` or using context manager which I did.
@asottile, please, take a look.

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+def normalize_content(content: bytes) -> bytes:
+    return b'\n'.join(
+        filter(
+            None,
+            map(
+                methodcaller('strip'),
+                content.splitlines(),
+            ),
+        ),
+    )
```
asottile, line 23: don't use `map` / `filter` / `reduce` / `methodcaller` -- the equivalent list comprehension is 100% more readable

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def find_destroyed_symlinks(autofix: bool) -> Sequence[bytes]:
+    destroyed_links = []
+    for line in check_output(
+        ['git', 'status', '--porcelain=v2', '-z'],
+    ).split(b'\0'):
+        splitted = line.split(b' ')
+        if splitted and splitted[0] == ORDINARY_CHANGED_ENTRIES_MARKER:
+            # variable names are taken from
```
asottile, line 33: these variable names are terrible, I don't particularly care if they match upstream let's make them readable

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
+import argparse
+import sys
+from operator import methodcaller
+from subprocess import check_call
+from subprocess import check_output
```
asottile, line 5: the rest of the project prefers not to `from` import from the standard library

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+                content.splitlines(),
+            ),
+        ),
+    )
+
+
+def find_destroyed_symlinks(autofix: bool) -> Sequence[bytes]:
+    destroyed_links = []
+    for line in check_output(
+        ['git', 'status', '--porcelain=v2', '-z'],
```
asottile, line 29: https://github.com/asottile/scratch/wiki/Prefer-tuples-to-lists

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+    ).split(b'\0'):
+        splitted = line.split(b' ')
```
asottile, line 31: is this safe to space split? does `.split()` do the same?

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+            ),
+        ),
+    )
+
+
+def find_destroyed_symlinks(autofix: bool) -> Sequence[bytes]:
+    destroyed_links = []
+    for line in check_output(
+        ['git', 'status', '--porcelain=v2', '-z'],
+    ).split(b'\0'):
```
asottile, line 30: there's a `zsplit` function that might be helpful, not sure if there's a copy here in `pre_commit_hooks/util.py` -- if not take the one from https://github.com/pre-commit/pre-commit/blob/3fa9d3d7588fa3aee37e04799ad36d2a0fe32e5a/pre_commit/git.py#L19-L24

that should remove the need for checking `if splitted` below

actually, looks like there's an implementation in `check-executables-have-shebangs` -- please move that into the util module and it's tests (if applicable)

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+                    # if old and new hashes are equal, it's not needed to check
+                    # anything more, we've found a destroyed symlink for sure
+                    found_destroyed_link = True
+                else:
+                    # if old and new hashes are *not* equal, it doesn't mean
+                    # that everything is OK - new file may be altered
+                    # by something like trailing-whitespace and/or
+                    # mixed-line-ending hooks so we need to go deeper
+                    index_size = int(
+                        check_output(['git', 'cat-file', '-s', hI]).strip(),
```
asottile, line 53: there's a `cmd_output` helper in `pre_commit_hooks/util.py` that probably makes most of this easier to read

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+                    )
+                    # Most filesystems limit path length to 4096 bytes.
```
asottile, line 55: windows allows paths longer than this with UNC naming

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,108 @@
...
+                        index_content = normalize_content(
+                            check_output(['git', 'cat-file', '-p', hI]),
+                        )
+                        found_destroyed_link = head_content == index_content
+                if found_destroyed_link:
+                    destroyed_links.append(path)
+                    if autofix:
+                        check_call([
+                            'git',
+                            'update-index',
```
asottile, line 76: pre-commit's stance is to never modify the staging area, I believe this does that so it can't be allowed -- printing this command seems fine though

.pre-commit-hooks.yaml
```diff
@@ -100,6 +100,13 @@
     entry: debug-statement-hook
     language: python
     types: [python]
+-   id: destroyed-symlinks
+    name: Detect Destroyed Symlinks
+    description: Detects symlinks which are changed to regular files with a content of a path which that symlink was pointing to.
+    entry: destroyed-symlinks
+    language: python
+    types: [file]
+    pass_filenames: false
```
asottile, line 109: this should still take filenames and filter the results to that -- that way during merge conflict resolution if someone else makes a mistake this doesn't blame the user performing the merge

tests/destroyed_symlinks_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+TEST_SYMLINK = 'test_symlink'
+
+
+@pytest.fixture
+def repo_with_destroyed_symlink(tmpdir):
+    source_repo = tmpdir.join('src')
+    os.makedirs(source_repo, exist_ok=True)
+    test_repo = tmpdir.join('test')
+    with source_repo.as_cwd():
```
asottile, line 19: note that you can use `git -C source_repo ...` instead of `as_cwd()`

pre_commit_hooks/destroyed_symlinks.py
asottile: what does this function do also? if it's to normalize newline endings it is not correct

m-khvoinitsky: > what does this function do also? if it's to normalize newline endings it is not correct

Not exactly. Apart from normalizing newline(s) and removing them it should help to detect the problem in some weird but possible cases:
* ex-symlink was pointing to file with trailing whitespace(s) at the end
* ex-symlink was pointing to file with newlines and whitespaces before newlines

and the content of ex-symlink was modified by end-of-file-fixer and/or mixed-line-ending and/or trailing-whitespace hooks.
Yes, false-positives are possible because of that but they are extremely unlikely, much more unlikely than the weird symlinks themselves.

asottile: do those scenarios actually happen?

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: I guess both variants are safe, I'll change it to plain `.split()`.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Actually, update-index is redundant here and even harmful — if the file was modified by some other hooks, symlinks would be broken. Instead, I should just unstage the file. Is this allowed?

asottile: I would just leave it and print a command for the user to run

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: I don't think so. But at least added newline at the end is practically possible so, if we aren't going to support weird filenames, at least plain `.rstrip()` is required.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Well, there definitely should be a limit in order to avoid wasting RAM in case when symlink was deliberately replaced with big file, which, on the other hand is not good idea in git so I don't know what is better.

tests/destroyed_symlinks_test.py
m-khvoinitsky: I know but this way it's a bit more readable IMO.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Even if this is opt-in feature (which it is)? May be rename `--autofix` to `--unstage`?

asottile: there's kinda two problems:
1. pre-commit won't notice the differences
2. it might get auto committed when wrong

if there's a way to modify the file to look like it's the symlink again in the unstaged portion that would be best

pre_commit_hooks/destroyed_symlinks.py
asottile: you can probably limit the read to the original symlink destination filename's length (that's how windows usually represents these on disk anyway iirc)

m-khvoinitsky: > you can probably limit the read to the original symlink destination filename's length

Plus a few bytes for various newlines. That's a good idea.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Oh, no, it's important to split by spaces because at the end there is filename which is rejoined later. If you have filenames with tabs or (more common) with more than one space in a row, you will have corrupted filename after rejoin.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Initially, I used `bytes` everywhere, so I've added support for `bytes` to `zsplit()`. However, after I've started using filenames passed as arguments, It became more tricky because `sys.argv` is `List[str]`. So, I gave up on using `bytes` but didn't removed the support from `zsplit()` — may be it would be useful someday.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky:
```
    kwargs.setdefault('stderr', subprocess.PIPE)
```
This eats stderr output which is probably not the best idea.

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: I've removed `--autofix`, now only the message with an instruction is displayed.

m-khvoinitsky: > if there's a way to modify the file to look like it's the symlink again in the unstaged portion that would be best

If it was possible, git would have done this. That's a problem.

pre_commit_hooks/destroyed_symlinks.py
asottile: "might be used one day" is known as code debt and maintenance burden -- please revert that if it's not being used

also a function taking both `str` and `bytes` is a wart, there should be `zsplit` and `zsplit_b` as separate functions

pre_commit_hooks/destroyed_symlinks.py
asottile: that's just the default, you can configure it otherwise.  and actually it's probably better as a pipe so the errors only display when the tool breaks

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: > also a function taking both str and bytes is a wart, there should be zsplit and zsplit_b as separate functions

Yes and no. If a function would automatically convert one to another, try to guess encoding (or assume ASCII) and so on, basically everything what was common in python2, then yes, that's not good.
In my case, it's more like a generic function: if you pass it bytes, it will return list of bytes and this even checked statically by mypy and various language servers so you won't mess one with another.

Anyway, I agree about dead code so I've removed it.

asottile: mmmm, yes and no but mostly no -- your function wasn't generic, [it manually checked the types](https://github.com/pre-commit/pre-commit-hooks/commit/82fe00463b4265fe2c419ba65ae645a2a647310d#) 😉

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+def find_destroyed_symlinks(files: Sequence[str]) -> Sequence[str]:
+    destroyed_links: List[str] = []
+    if not files:
+        return destroyed_links
+    for line in zsplit(
+        cmd_output(
+            *itertools.chain(
+                ('git', 'status', '--porcelain=v2', '-z', '--'),
+                files,
+            ),
```
asottile, line 30: this can be:

```python
cmd_output('git', 'status', '--porcelain=v2', '-z', '--', *files)
```

README.md
```diff
@@ -82,6 +82,15 @@ Attempts to load all yaml files to verify syntax.
 #### `debug-statements`
 Check for debugger imports and py37+ `breakpoint()` calls in python source.
 
+#### `destroyed-symlinks`
+Detects symlinks which are changed to regular files with a content of a path which that symlink was pointing to.
+This usually happens on Windows in case when user without a permission for creating symlinks clones repository with symlinks.
+The following argument is available:
+- `--autofix` - unstage detected broken symlinks so they won't be commited.
```
asottile, line 89: I believe this option is gone now

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+                hash_HEAD,
+                hash_index,
+                *path_splitted,
+            ) = splitted
+            path = ' '.join(path_splitted)
+            if all((
+                mode_HEAD == PERMS_LINK,
+                mode_index != PERMS_LINK,
+                mode_index != PERMS_NONEXIST,
+            )):
```
asottile, line 50: `all((` here is going to be slower than putting `and` to join these conditions

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+from pre_commit_hooks.util import cmd_output
+from pre_commit_hooks.util import zsplit
+
+ORDINARY_CHANGED_ENTRIES_MARKER = '1'
+PERMS_LINK = '120000'
+PERMS_NONEXIST = '000000'
+
+
+def normalize_content(content: bytes) -> bytes:
+    return content.rstrip()
```
asottile, line 18: I'd probably just inline this

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+    args = parser.parse_args(argv)
+    destroyed_links = find_destroyed_symlinks(files=args.filenames)
+    if destroyed_links:
+        print('Destroyed symlinks:')
+        for destroyed_link in destroyed_links:
+            print(f'- {destroyed_link}')
+        print('You should unstage affected files:')
+        print(
+            ' git reset HEAD -- {}'.format(
+                ' '.join(map(shlex.quote, destroyed_links)),
```
asottile, line 95: please don't use `map` / `filter` / `reduce` -- a list comprehension is more readable

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+            ),
+        )
+        print(
+            'And retry commit. As a long term solution '
+            'you may try to explicitly tell git that your '
+            'envionment does not support symlinks:',
```
asottile, line 101: environment

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+        print(' git config --local core.symlinks false')
```
asottile, line 103: iirc `--local` is the default for this command

pre_commit_hooks/util.py
```diff
@@ -13,7 +15,11 @@ def added_files() -> Set[str]:
     return set(cmd_output(*cmd).splitlines())
 
 
-def cmd_output(*cmd: str, retcode: Optional[int] = 0, **kwargs: Any) -> str:
+def cmd_output(
+    *cmd: Union[str, bytes],
```
asottile, line 19: mixing here used to have problems on windows, I'm not sure if it's still a problem in python3 though

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: Tastes differ.

pre_commit_hooks/util.py
m-khvoinitsky: Reverted. It was needed in the previous version where everything was `bytes`. I don't have Windows nearby to check at the moment, however, tests weren't failing.
They use [os.fsdecode](https://docs.python.org/3/library/os.html#os.fsdecode) ([subprocess.py#L561](https://github.com/python/cpython/blob/1642c0ef750f96664a98cadb09301d492098d2fb/Lib/subprocess.py#L561)) which, quoting the doc, _decode the path-like filename from the filesystem encoding with 'surrogateescape' error handler, or 'strict' on Windows; return `str` unchanged_ (good bad example of mixing `str` and `bytes`).

pre_commit_hooks/destroyed_symlinks.py
```diff
@@ -34,9 +34,9 @@ def find_destroyed_symlinks(files: Sequence[str]) -> Sequence[str]:
             ) = splitted
             path = ' '.join(path_splitted)
             if (
-                mode_HEAD == PERMS_LINK
-                and mode_index != PERMS_LINK
-                and mode_index != PERMS_NONEXIST
+                mode_HEAD == PERMS_LINK and
+                mode_index != PERMS_LINK and
+                mode_index != PERMS_NONEXIST
```
m-khvoinitsky, line 39: https://www.flake8rules.com/rules/W504.html
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

asottile: you don't need to lecture me, I maintain both pycodestyle and flake8 :)

asottile: if you're curious about the history here though, pep8 has gone back and forth on this twice now and I'm not interested in the churn -- plus I think it's much more important that the conditions line up than the operators

and fortunately you don't need to have a discussion about it at all because the formatters enforce it!

m-khvoinitsky: Nice :)
Anyway, for some reason, locally pre-commit didn't complain about this part.

asottile: weird!  it might've been due to a rebase (since `.git/hooks/pre-commit` doesn't fire on a `git rebase --continue`)?

m-khvoinitsky: > plus I think it's much more important that the conditions line up than the operators

That's why I used `all(...)` in the first place.

> and fortunately you don't need to have a discussion about it at all because the formatters enforce it!

No offences here, I've started discussion mainly because I was surprised with different behaviour between local and CI versions. The reason seems to be the fact that hook uses system autopep8 and, apparently, we have different versions.

asottile (approved):

pre_commit_hooks/destroyed_symlinks.py
m-khvoinitsky: > weird! it might've been due to a rebase (since .git/hooks/pre-commit doesn't fire on a git rebase --continue)?

I've launched `tox` manually (which launches pre-commit for all project).

My versions: autopep8 1.5.4 (pycodestyle: 2.6.0)

asottile: there's probably some global configuration you have -- it doesn't use system `autopep8` (that's most of the point of `pre-commit` in that you don't have to install things globally!)

m-khvoinitsky:
Thanks!

PR #514, opened 2020-08-27
Add a way to do case-insensitive sorting via file-contents-sorter.

Julian:
Does nothing fancy (no case folding).

But does what I need in my case.

Comments welcome of course.

### DIFF
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -10,6 +10,8 @@ this hook on that file should reduce the instances of git merge
 conflicts and keep the file nicely ordered.
 """
 import argparse
+from typing import Any
+from typing import Callable
 from typing import IO
 from typing import Optional
 from typing import Sequence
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/file_contents_sorter.py
```diff
@@ -37,13 +40,20 @@ def sort_file_contents(f: IO[bytes]) -> int:
 def main(argv: Optional[Sequence[str]] = None) -> int:
     parser = argparse.ArgumentParser()
     parser.add_argument('filenames', nargs='+', help='Files to sort')
+    parser.add_argument(
+        '--ignore-case',
+        action='store_const',
+        const=bytes.lower,
+        default=None,
+        help='fold lower case to upper case characters',
```
Julian, line 48: (Argument name and help happen to be grabbed from sort(1) from coreutils, though I didn't also add `-f`, since that's just confusing... can though if desired.)

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -18,9 +18,12 @@
 FAIL = 1
 
 
-def sort_file_contents(f: IO[bytes]) -> int:
+def sort_file_contents(f: IO[bytes], key : Optional[callable]) -> int:
```
asottile, line 21: mypy points out this should be `Optional[Callable[[bytes], Any]]`

Julian: Fixed.

Julian:
pfft so did I.

Clearly pre-commit.ci should just be the PR submitter and maintainer racing to run pre-commit.

PR #519, opened 2020-09-24
Fix #518, provide --enforce-all option to check_added_large_files

### LINKED ISSUE #518: check-added-large-files does not check files when run in a CI loop
check-added-large-files works as expected when executed as a pre-commit hook, but when executed in a CI loop via run --all-files it does not diagnose over size files.  This appears to be because the list of files provided on the command line is filtered to remove files that are not reported as staged for addition by git diff.  Hence if a developer chooses not to install pre-commit as a git hook and adds an over size file, this will not subsequently be picked up by pre-commit running in CI.

mshawcroft:
The --enforce-all option when provided ensures that all files passed
on the command line are checked against the size limit.  Default
behaviour remains unchanged.

### DIFF
--- a/pre_commit_hooks/check_added_large_files.py
+++ b/pre_commit_hooks/check_added_large_files.py
@@ -21,11 +21,20 @@ def lfs_files() -> Set[str]:
     return set(json.loads(lfs_ret)['files'])
 
 
-def find_large_added_files(filenames: Sequence[str], maxkb: int) -> int:
+def find_large_added_files(
+        filenames: Sequence[str],
+        maxkb: int,
+        *,
+        enforce_all: bool = False,
+) -> int:
     # Find all added files that are also in the list of files pre-commit tells
     # us about
     retv = 0
-    for filename in (added_files() & set(filenames)) - lfs_files():
+    filenames_filtered = set(filenames) - lfs_files()
+    if not enforce_all:
+        filenames_filtered &= added_files()
+
+    for filename in filenames_filtered:
         kb = int(math.ceil(os.stat(filename).st_size / 1024))
         if kb > maxkb:
             print(f'{filename} ({kb} KB) exceeds {maxkb} KB.')
@@ -41,12 +50,20 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
         help='Filenames pre-commit believes are changed.',
     )
     parser.add_argument(
+        '--enforce-all', action='store_true',
+        help='Enforce all files are checked, not just staged files.',
+    )
+    parser.add_argument(
         '--maxkb', type=int, default=500,
         help='Maxmimum allowable KB for added files',
     )
-
     args = parser.parse_args(argv)
-    return find_large_added_files(args.filenames, args.maxkb)
+
+    return find_large_added_files(
+        args.filenames,
+        args.maxkb,
+        enforce_all=args.enforce_all,
+    )
 
 
 if __name__ == '__main__':
--- a/tests/check_added_large_files_test.py
+++ b/tests/check_added_large_files_test.py
@@ -40,6 +40,17 @@ def test_add_something_giant(temp_git_dir):
         assert find_large_added_files(['f.py'], 10) == 0
 
 
+def test_enforce_all(temp_git_dir):
+    with temp_git_dir.as_cwd():
+        temp_git_dir.join('f.py').write('a' * 10000)
+
+        # Should fail, when not staged with enforce_all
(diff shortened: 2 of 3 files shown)


pre_commit_hooks/check_added_large_files.py
```diff
@@ -21,11 +21,19 @@ def lfs_files() -> Set[str]:
     return set(json.loads(lfs_ret)['files'])
 
 
-def find_large_added_files(filenames: Sequence[str], maxkb: int) -> int:
+def find_large_added_files(
+    filenames: Sequence[str], maxkb: int,
+    enforce_all: Optional[bool] = False,
```
asottile, line 26: `Optional[bool]` means `Union[bool, None]` -- you don't want `Optional` here, just `bool`

I would make this a keyword-only argument to enforce that the calling construct is readable -- otherwise this reads like a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

(not a big deal, but I usually double-indent parameters to differentiate them from the function body, but yeah I know I don't have a formatter -- I wouldn't have mentioned this if I wasn't commenting on the PR for other reasons)

pre_commit_hooks/check_added_large_files.py
```diff
@@ -21,11 +21,19 @@ def lfs_files() -> Set[str]:
...
+) -> int:
     # Find all added files that are also in the list of files pre-commit tells
     # us about
     retv = 0
-    for filename in (added_files() & set(filenames)) - lfs_files():
+    filename_set = set(filenames)
+    if not enforce_all:
+        filename_set &= added_files()
+        filename_set -= lfs_files()
```
asottile, line 34: this should still exclude lfs_files even with `--enforce-all`, they aren't relevant to the repository size

pre_commit_hooks/check_added_large_files.py
```diff
@@ -40,13 +48,20 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
         'filenames', nargs='*',
         help='Filenames pre-commit believes are changed.',
     )
+    parser.add_argument(
+        '--enforce-all', action='store_true',
+        help='Enforce all files are checked.',
```
asottile, line 53: maybe mention `(not just staged files)`

tests/check_added_large_files_test.py
```diff
@@ -39,6 +39,19 @@ def test_add_something_giant(temp_git_dir):
...
+        assert find_large_added_files(['f.py'], 0, False) == 1
+
+        # File is committed hence no longer staged
+        cmd_output('git', 'commit', '-m', 'log')
+
+        # Should fail, when not staged with enforce_all
+        assert find_large_added_files(['f.py'], 0, True) == 1
+
+        # Should pass, when not staged without enforce_all
+        assert find_large_added_files(['f.py'], 0, False) == 0
```
asottile, line 53: these should be separate tests since they are validating different concerns

tests/check_added_large_files_test.py
```diff
@@ -80,8 +93,10 @@ def test_allows_gitlfs(temp_git_dir, monkeypatch):  # pragma: no cover
         temp_git_dir.join('f.py').write('a' * 10000)
         cmd_output('git', 'lfs', 'track', 'f.py')
         cmd_output('git', 'add', '--', '.')
-        # Should succeed
+        # Without --enforce-all large files on git lfs should succeed
         assert main(('--maxkb', '9', 'f.py')) == 0
+        # With --enforce-all large files on git lfs should fail
+        assert main(('--enforce-all', '--maxkb', '9', 'f.py')) == 1
```
asottile, line 99: this'll change as discussed above

tests/check_added_large_files_test.py
```diff
@@ -97,3 +112,4 @@ def test_moves_with_gitlfs(temp_git_dir, monkeypatch):  # pragma: no cover
         # Now move it and make sure the hook still succeeds
         cmd_output('git', 'mv', 'a.bin', 'b.bin')
         assert main(('--maxkb', '9', 'b.bin')) == 0
+        assert main(('--enforce-all', '--maxkb', '9', 'b.bin')) == 1
```
asottile, line 115: I would test this in a separate test as well

pre_commit_hooks/check_added_large_files.py
mshawcroft: Thanks for the link on api design.

asottile (approved):

PR #522, opened 2020-10-05
Attempt to add fixing of BOMs

jgowdy:
I'm attempting to add automatic fixing of BOMs

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -17,8 +17,8 @@
     language: python
     types: [python]
 -   id: check-byte-order-marker
-    name: Check for byte-order marker
-    description: Forbid files which have a UTF-8 byte-order marker
+    name: 'check BOM - deprecated: use fix-byte-order-marker'
+    description: forbid files which have a UTF-8 byte-order marker
     entry: check-byte-order-marker
     language: python
     types: [text]
@@ -131,6 +131,12 @@
     entry: file-contents-sorter
     language: python
     files: '^$'
+-   id: fix-byte-order-marker
+    name: fix UTF-8 byte order marker
+    description: removes UTF-8 byte order marker
+    entry: fix-byte-order-marker
+    language: python
+    types: [text]
 -   id: fix-encoding-pragma
     name: Fix python encoding pragma
     language: python
--- a/pre_commit_hooks/fix_byte_order_marker.py
+++ b/pre_commit_hooks/fix_byte_order_marker.py
@@ -1,0 +1,30 @@
+import argparse
+from typing import Optional
+from typing import Sequence
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*', help='Filenames to check')
+    args = parser.parse_args(argv)
+
+    retv = 0
+
+    for filename in args.filenames:
+        with open(filename, 'rb') as f_b:
+            bts = f_b.read(3)
+
+        if bts == b'\xef\xbb\xbf':
+            with open(filename, newline='', encoding='utf-8-sig') as f:
+                contents = f.read()
+            with open(filename, 'w', newline='', encoding='utf-8') as f:
+                f.write(contents)
+
+            print(f'{filename}: removed byte-order marker')
+            retv = 1
+
+    return retv
+
+
+if __name__ == '__main__':
+    exit(main())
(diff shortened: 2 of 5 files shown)


jgowdy:
Copied the remove_bom_inplace function from https://www.stefangordon.com/remove-bom-mark-from-text-files-in-python/

pre_commit_hooks/byte_order_marker.py
```diff
@@ -0,0 +1,51 @@
...
+                if args.fix == 'no':
+                    print(f'{filename}: Found byte-order marker')
+                else:
+                    remove_bom_inplace(filename)
+                    print(f'{filename}: Removed byte-order marker')
+                retv = 1
+
+    return retv
+
+def remove_bom_inplace(path):
```
asottile, line 32: sorry we can't take this code because it is not properly licensed, also it is unnecessarily complicated

an easier way to do this is:

```python
with open(..., newline='', encoding='utf-8-sig') as f:
    contents =f.read()

with open(..., newline='', encoding='utf-8') as f:
    f.write(contents)
```

setup.cfg
```diff
@@ -28,10 +28,10 @@ python_requires = >=3.6.1
 
 [options.entry_points]
 console_scripts =
+    byte-order-marker = pre_commit_hooks.byte_order_marker:main
     check-added-large-files = pre_commit_hooks.check_added_large_files:main
     check-ast = pre_commit_hooks.check_ast:main
     check-builtin-literals = pre_commit_hooks.check_builtin_literals:main
-    check-byte-order-marker = pre_commit_hooks.check_byte_order_marker:main
```
asottile, line 34: this is a breaking change

asottile: and not reflected in `.pre-commit-hooks.yaml`

asottile (commented):
I think we should deprecate check-byte-order-marker, but we cannot remove it in this PR

pre_commit_hooks/byte_order_marker.py
asottile: oh, you have to actually fill in the `...`s 🤦

jgowdy:
How would I mark the existing check-byte-order-marker check as deprecated?

jgowdy:
Sorry for the churn, I don't write much Python

jgowdy:
For some reason the CacheBeta step of CI is failing / timing out

pre_commit_hooks/byte_order_marker.py
jgowdy: Yeah... not enough coffee at that point

jgowdy:
This passed in this test run, and I'm now rebasing out my "bump" commits, but hopefully it doesn't fail again.

https://asottile.visualstudio.com/asottile/_build/results?buildId=4599&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d

jgowdy:
Okay, I think this is good now.

rohitjha (approved):

jgowdy:
Would you mind adding a hacktoberfest-accepted label to this PR?  I didn't make this change specifically for hacktoberfest, I need it, but I am participating in hacktoberfest.  Thanks!

asottile:
yeah sorry I've been slow about this -- I'll take a look when I get the chance 👍

.pre-commit-hooks.yaml
```diff
@@ -17,8 +17,8 @@
     language: python
     types: [python]
 -   id: check-byte-order-marker
-    name: Check for byte-order marker
-    description: Forbid files which have a UTF-8 byte-order marker
+    name: 'check BOM - deprecated: use fix-byte-order-marker'
```
asottile, line 20: I shortened the name here so it doesn't wrap -- I should probably add a test for this

.pre-commit-hooks.yaml
```diff
@@ -131,6 +131,12 @@
     entry: file-contents-sorter
     language: python
     files: '^$'
+-   id: fix-byte-order-marker
```
asottile, line 134: I renamed this to `fix-byte-order-marker` so it's more clear what it does

pre_commit_hooks/fix_byte_order_marker.py
```diff
@@ -0,0 +1,30 @@
+import argparse
+from typing import Optional
+from typing import Sequence
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*', help='Filenames to check')
```
asottile, line 8: I removed the `--fix=no` -- I'd rather have formatters-only

pre_commit_hooks/fix_byte_order_marker.py
```diff
@@ -0,0 +1,30 @@
...
+    args = parser.parse_args(argv)
+
+    retv = 0
+
+    for filename in args.filenames:
+        with open(filename, 'rb') as f_b:
+            bts = f_b.read(3)
```
asottile, line 15: there was a leaking file descriptor here, I fixed that

asottile (approved):

PR #524, opened 2020-10-25
Add --unique arg to file_contents_sorter

danielhoherd:
Allow file_contents_sorter to unique a file via the new --unique arg.

### DIFF
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -13,6 +13,7 @@ import argparse
 from typing import Any
 from typing import Callable
 from typing import IO
+from typing import Iterable
 from typing import Optional
 from typing import Sequence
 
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/file_contents_sorter.py
```diff
@@ -52,13 +59,22 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
         default=None,
         help='fold lower case to upper case characters',
     )
+    parser.add_argument(
+        '--unique',
+        action='store_const',
+        const=True,
+        default=False,
```
asottile, line 66: delete these three lines and use `action='store_true'`

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -23,12 +23,19 @@
 def sort_file_contents(
     f: IO[bytes],
     key: Optional[Callable[[bytes], Any]],
+    unique: bool,
```
asottile, line 26: this should be keyword only, otherwise it's a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap)

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -23,12 +23,19 @@
...
+    if unique:
+        after = sorted(
+            {line.strip(b'\n\r') for line in before if line.strip()},
+            key=key,
+        )
+    else:
+        after = sorted(
+            (line.strip(b'\n\r') for line in before if line.strip()),
+            key=key,
+        )
```
asottile, line 38: maybe this is a little easier to read?
```python
lines: Iterable[bytes] = (line.rstrip(b'\n\r') for line in before if line.strip())
if unique:
    lines = set(lines)
after = sorted(lines, key=key)
```

pre_commit_hooks/file_contents_sorter.py
asottile: this didn't get fixed, here's a vid where I talk about [all the parameter types](https://youtu.be/aKCfCmSggPg)

asottile (approved):

PR #530, opened 2020-11-16
Add --additional-domain to check-vcs-permalinks

youngminz:
I have added an option to inspect the source code of the additional domain, especially useful on GitHub Enterprise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ Attempts to load all TOML files to verify syntax.
 
 #### `check-vcs-permalinks`
 Ensures that links to vcs websites are permalinks.
+  - `--additional-github-domain DOMAIN` - Add check for specified domain.
+    Can be repeated multiple times.  for example, if your company uses
+    GitHub Enterprise you may use something like
+    `--additional-github-domain github.example.com`
 
 #### `check-xml`
 Attempts to load all xml files to verify syntax.
--- a/pre_commit_hooks/check_vcs_permalinks.py
+++ b/pre_commit_hooks/check_vcs_permalinks.py
@@ -1,35 +1,50 @@
 import argparse
 import re
 import sys
+from typing import List
 from typing import Optional
+from typing import Pattern
 from typing import Sequence
 
 
-GITHUB_NON_PERMALINK = re.compile(
-    br'https://github.com/[^/ ]+/[^/ ]+/blob/master/[^# ]+#L\d+',
-)
+def _get_pattern(domain: str) -> Pattern[bytes]:
+    regex = rf'https://{domain}/[^/ ]+/[^/ ]+/blob/master/[^# ]+#L\d+'
+    return re.compile(regex.encode())
 
 
-def _check_filename(filename: str) -> int:
+def _check_filename(filename: str, patterns: List[Pattern[bytes]]) -> int:
     retv = 0
     with open(filename, 'rb') as f:
         for i, line in enumerate(f, 1):
-            if GITHUB_NON_PERMALINK.search(line):
-                sys.stdout.write(f'{filename}:{i}:')
-                sys.stdout.flush()
-                sys.stdout.buffer.write(line)
-                retv = 1
+            for pattern in patterns:
+                if pattern.search(line):
+                    sys.stdout.write(f'{filename}:{i}:')
+                    sys.stdout.flush()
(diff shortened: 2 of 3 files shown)


pre_commit_hooks/check_vcs_permalinks.py
```diff
@@ -25,11 +24,23 @@ def _check_filename(filename: str) -> int:
...
     retv = 0
+
+    github_non_permalink = re.compile(NON_PERMALINK % b'github.com')
     for filename in args.filenames:
-        retv |= _check_filename(filename)
+        retv |= _check_filename(filename, github_non_permalink)
+
+    if args.github_enterprise_domain:
+        github_enterprise_non_permalink = re.compile(
+            NON_PERMALINK % args.github_enterprise_domain.encode(),
```
asottile, line 40: please don't use %s, if anything use .format() or f-strings -- thanks!

I notice this is a bytestring which doesn't support .format() or f-strings, but you can change it to not be a bytestring and then do .encode() afterwards as you've done for the domain

you can probably also inline the string as well

pre_commit_hooks/check_vcs_permalinks.py
```diff
@@ -25,11 +24,23 @@ def _check_filename(filename: str) -> int:
...
+        )
+        for filename in args.filenames:
+            retv |= _check_filename(filename, github_enterprise_non_permalink)
```
asottile, line 43: this is going to re-parse twice, maybe add an additional argument to `_check_filename` such that it only runs over each file twice?  this will also help with output ordering as well

tests/check_vcs_permalinks_test.py
```diff
@@ -33,3 +33,43 @@ def test_failing(tmpdir, capsys):
...
+def test_passing_with_github_enterprise(tmpdir):
+    f = tmpdir.join('f.txt')
+    f.write_binary(
+        # permalinks are ok
+        b'https://example.com/asottile/test/blob/649e6/foo%20bar#L1\n'
+        # links to files but not line numbers are ok
+        b'https://example.com/asottile/test/blob/master/foo%20bar\n'
+        # regression test for overly-greedy regex
+        b'https://example.com/ yes / no ? /blob/master/foo#L1\n',
+    )
```
asottile, line 47: you can just use a simpler test, no need to re-test everything again

README.md
```diff
@@ -66,6 +66,7 @@ Attempts to load all TOML files to verify syntax.
 
 #### `check-vcs-permalinks`
 Ensures that links to vcs websites are permalinks.
+  - `--github-enterprise-domain DOMAIN` - Add check for specified GitHub Enterprise domain.
```
asottile, line 69: maybe `--additional-domain` is simpler? it doesn't necessarily have to be github enterprise, just something with similar url structure

pre_commit_hooks/check_vcs_permalinks.py
youngminz: Since the `NON_PERMALINK` is used only inside the function that converts the domain to a regular expression, I changed it to inline and used the f-string.

pre_commit_hooks/check_vcs_permalinks.py
youngminz: Rather than increasing the logic branch, using a list can make it simpler, so I tried using a list. I tried modifying action='append' to allow users to add more than one domain. How about this?

tests/check_vcs_permalinks_test.py
youngminz: Since the logic does not change even if the user specifies additional domains to check, I added a case that fails. Are there any more test cases to add?

README.md
youngminz: Applied it!

asottile (approved):
since it looks like your company is using pre-commit, maybe consider [sponsoring](https://github.com/sponsors/asottile) (I'm currently unemployed!)

youngminz:
Thanks to pre-commit, our team can reduce unnecessary convention discussions and have more productive business logic discussions!  I hope my sponsor is helpful to you.

asottile:
> 
> 
> Thanks to pre-commit, our team can reduce unnecessary convention discussions and have more productive business logic discussions! I hope my sponsor is helpful to you.

<3 thanks!  I really appreciate it!

PR #546, opened 2021-01-15
Add missing ruamel.yaml test dependency

l0b0:
Fixes this issue:

```
pre_commit_hooks/check_yaml.py:8: in <module>
    import ruamel.yaml
E   ModuleNotFoundError: No module named 'ruamel'
```


asottile:
it's in setup.cfg already

PR #571, opened 2021-03-14
Update `pretty-format-json` to write to stdout with utf-8 encoding

### LINKED ISSUE #570: `pretty-format-json` says the JSON file is not valid when it is (Windows-specific?)
Here's a full log from GitHub runner presenting this issue:
https://github.com/jack1142/Red-DiscordBot/runs/2107108269?check_suite_focus=true

I tried to make a simple repro on GH Actions using Linux environment but I didn't get this error so I tried using the same environment I have this problem with locally - Windows - and then I was indeed able to reproduce it on the GH Actions runner.

`check-json` says the JSON file is valid and as far as I can tell, it is.

jack1142:
Fixes #570

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+import sys
 from difflib import unified_diff
 from typing import List
 from typing import Mapping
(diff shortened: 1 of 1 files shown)


jack1142:
If needed, I can rebase this to remove the noisy formatting commits.

pre_commit_hooks/pretty_format_json.py
```diff
@@ -116,10 +117,10 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
...
-                    print(
-                        get_diff(contents, pretty_contents, json_file),
-                        end='',
+                    diff_output = get_diff(
+                        contents, pretty_contents, json_file,
                     )
+                    sys.stdout.buffer.write(diff_output.encode('utf-8'))
 
                 status = 1
         except ValueError:
```
asottile, line 126: the other problem with how this is written is there's wayyy too much in this `try:` `except:` -- part of the fix should be to narrow the scope to only the raisable code (which was hiding this bug)

pre_commit_hooks/pretty_format_json.py
```diff
@@ -111,24 +112,24 @@ def main(argv: Optional[Sequence[str]] = None) -> int:
...
             return 1
 
+        if contents != pretty_contents:
+            if args.autofix:
+                _autofix(json_file, pretty_contents)
+            else:
+                diff_output = get_diff(
+                    contents, pretty_contents, json_file,
+                )
+                sys.stdout.buffer.write(diff_output.encode('utf-8'))
```
asottile, line 129: UTF-8 is the default encoding in all versions supported, this can be `.encode()`

asottile (approved):

PR #582, opened 2021-04-06
Disallow any branch name in checks_vcs_permalinks

### LINKED ISSUE #581: `check-vcs-permalinks` - Add support for different branch names
Not all repositories use `master` for their branch name so it would be great if this hook could support other branch names.

Preferably, this would match any blob link that doesn't contain a full hash (so *only* matching 0-f with len of 40 or 64) but if you don't consider that to be a good solution, then being able to specify a list of branches to match would also work, though it would be a lot more likely to lead to non-permalinks not getting matched.

jack1142:
Fixes #581

### DIFF
--- a/pre_commit_hooks/check_vcs_permalinks.py
+++ b/pre_commit_hooks/check_vcs_permalinks.py
@@ -8,7 +8,10 @@ from typing import Sequence
 
 
 def _get_pattern(domain: str) -> Pattern[bytes]:
-    regex = rf'https://{domain}/[^/ ]+/[^/ ]+/blob/master/[^# ]+#L\d+'
+    regex = (
+        rf'https://{domain}/[^/ ]+/[^/ ]+/blob/'
+        r'(?![a-fA-F0-9]{4,64}/)([^/. ]+)/[^# ]+#L\d+'
+    )
     return re.compile(regex.encode())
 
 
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/check_vcs_permalinks.py
```diff
@@ -7,8 +7,11 @@
 from typing import Sequence
 
 
+SHA_REGEX = re.compile(rb'^[a-fA-F0-9]{4,64}$')
+
+
 def _get_pattern(domain: str) -> Pattern[bytes]:
-    regex = rf'https://{domain}/[^/ ]+/[^/ ]+/blob/master/[^# ]+#L\d+'
+    regex = rf'https://{domain}/[^/ ]+/[^/ ]+/blob/([^/. ]+)/[^# ]+#L\d+'
```
asottile, line 14: hmm I feel like this can be done with the pattern entirely (and not need to change the rest of the code) -- perhaps with lookaheads (atomic matches)?

asottile:
the title says "Allow", but specifically it's the opposite

pre_commit_hooks/check_vcs_permalinks.py
jack1142: Yeah, you're right, I unnecessarily avoided the use of lookahead. I reverted the changes and updated just the regex this time.

asottile (approved):

PR #583, opened 2021-04-07
Add gitleaks hook

rtluckie:


rtluckie:
@asottile no interest in addition of this hook?  I meant to open PR as WIP as I still need to add tests, but I though I'd check to see if there was any  interest.

asottile:
sorry, pre-commit/pre-commit-hooks is not accepting new hooks -- and especially not ones which require external installation (kinda violates the whole goals of the framework).  also pre-commit itself has full support for `golang` as a language so you should be able to write this hook with zero code (ideally by contributing a `.pre-commit-hooks.yaml` to gitleaks itself, or by writing a `local` hook with it in `additional_dependencies`)

PR #586, opened 2021-04-10
Check for diff3-style merge conflict artifacts.

christhekeele:
I noticed that [the test suite](https://github.com/pre-commit/pre-commit-hooks/blob/master/tests/check_merge_conflict_test.py#L49) checks for `diff3`-style merge conflicts, but the [actual hook](https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/check_merge_conflict.py#L8-L11) does not. Seems like an oversight?

### DIFF
--- a/pre_commit_hooks/check_merge_conflict.py
+++ b/pre_commit_hooks/check_merge_conflict.py
@@ -7,6 +7,7 @@ from typing import Sequence
 CONFLICT_PATTERNS = [
     b'<<<<<<< ',
     b'======= ',
+    b'||||||| ',
     b'=======\n',
     b'>>>>>>> ',
 ]
--- a/pre_commit_hooks/check_merge_conflict.py
+++ b/pre_commit_hooks/check_merge_conflict.py
@@ -7,8 +7,9 @@ from typing import Sequence
 CONFLICT_PATTERNS = [
     b'<<<<<<< ',
     b'======= ',
-    b'||||||| ',
     b'=======\n',
+    b'||||||| ',
+    b'|||||||\n',
     b'>>>>>>> ',
 ]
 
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/check_merge_conflict.py
```diff
@@ -7,6 +7,7 @@
 CONFLICT_PATTERNS = [
     b'<<<<<<< ',
     b'======= ',
+    b'||||||| ',
```
asottile, line 10: would you like to add a failing test for this?  the testsuite passes because that particular string is part of a larger string which contains the other strings

christhekeele: Sure! Think I can get to that later tonight

christhekeele: @asottile I gave this a go, but am encountering some friction with running tests locally (test dependencies are missing from requirements-dev.test and I'm getting unexplicable tmpfile errors). The build is also telling me that there is something incorrect about either my understanding of the how the tests work. So not pursuing for now, but leaving this open for others!

asottile: I think all you need to do is add to this list: https://github.com/pre-commit/pre-commit-hooks/pull/586/files#diff-6224b62678ea600729955336d2e3f48b0c90be48e8ea57dbefe54e7c00131532R106

are you using `tox` when testing locally?

christhekeele: A bit confused there, that commit... is where I added to that list? Then CI fails on [this test](https://github.com/pre-commit/pre-commit-hooks/pull/586/commits/0c6a81a69f43717f79028f9938ca0d9d68ba8eeb#diff-6224b62678ea600729955336d2e3f48b0c90be48e8ea57dbefe54e7c00131532R136-R138) that I'd expect to succeed.

I've never used `tox` before but I figured it out (tho still have odd tmpfile-caused test failures). Some dev-instruction that you should be using it, and how to use, may help with future contributors.

Speaking of which, thanks for the help!

asottile: yeah you shouldn't need to touch that test -- it's testing a different behaviour (when `--assume-in-merge` is passed it *should* check for those strings) -- I think you can revert your test changes and just add to the tests on line 106

PR #590, opened 2021-04-22
Add check for illegal Widnows names

### LINKED ISSUE #589: Add check for illegal Windows filenames
There are file and directory names which are legal in Git and most operating systems but illegal in Windows.  They should be avoided in case a developer happens to use Windows.

The following names are illegal as bare names as well as with extensions regardless of case for files or directories.  For instance `CoM6`, `cOm6.txt`, `com6.foo.jpg`, etc are all illegal.  Windows will refuse to create those files if you try to check them out.

From: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

> Do not use the following reserved names for the name of a file:
> 
> CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. Also avoid these names followed immediately by an extension; for example, NUL.txt is not recommended. For more information, see [Namespaces](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file).

ericfrederich:
Fixes #589

### DIFF
--- a/pre_commit_hooks/check_illegal_windows_names.py
+++ b/pre_commit_hooks/check_illegal_windows_names.py
@@ -1,0 +1,66 @@
+import argparse
+import os.path
+from typing import Iterable
+from typing import Iterator
+from typing import Optional
+from typing import Sequence
+from typing import Set
+
+from pre_commit_hooks.util import added_files
+
+
+def lower_set(iterable: Iterable[str]) -> Set[str]:
+    return {x.lower() for x in iterable}
+
+
+def parents(file: str) -> Iterator[str]:
+    file = os.path.dirname(file)
+    while file:
+        yield file
+        file = os.path.dirname(file)
+
+
+def directories_for(files: Set[str]) -> Set[str]:
+    return {parent for file in files for parent in parents(file)}
+
+
+# https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+ILLEGAL_NAMES = {
+    'CON',
+    'PRN',
+    'AUX',
+    'NUL',
+    *(f'COM{i}' for i in range(1, 10)),
+    *(f'LPT{i}' for i in range(1, 10)),
+}
+
+
+def find_illegal_windows_names(filenames: Sequence[str]) -> int:
+    relevant_files = set(filenames) | added_files()
+    relevant_files |= directories_for(relevant_files)
+    retv = 0
+
+    for filename in relevant_files:
+        root = os.path.basename(filename)
+        while '.' in root:
+            root, _ = os.path.splitext(root)
+        if root.lower() in lower_set(ILLEGAL_NAMES):
+            print(f'Illegal name {filename}')
+            retv = 1
+    return retv
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'filenames', nargs='*',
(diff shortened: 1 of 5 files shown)


pre_commit_hooks/check_illegal_windows_names.py
```diff
@@ -0,0 +1,66 @@
...
+        yield file
+        file = os.path.dirname(file)
+
+
+def directories_for(files: Set[str]) -> Set[str]:
+    return {parent for file in files for parent in parents(file)}
+
+
+# https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+ILLEGAL_NAMES = {
```
asottile, line 28: `frozenset((`

pre_commit_hooks/check_illegal_windows_names.py
```diff
@@ -0,0 +1,66 @@
...
+    relevant_files = set(filenames) | added_files()
+    relevant_files |= directories_for(relevant_files)
+    retv = 0
+
+    for filename in relevant_files:
+        root = os.path.basename(filename)
+        while '.' in root:
+            root, _ = os.path.splitext(root)
+        if root.lower() in lower_set(ILLEGAL_NAMES):
+            print(f'Illegal name {filename}')
```
asottile, line 48: Illegal windows filename: {filename}

pre_commit_hooks/check_illegal_windows_names.py
```diff
@@ -0,0 +1,66 @@
...
+    return retv
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        'filenames', nargs='*',
+        help='Filenames pre-commit believes are changed.',
+    )
+
```
asottile, line 59: can remove this newline

pre_commit_hooks/check_illegal_windows_names.py
```diff
@@ -0,0 +1,66 @@
...
+    args = parser.parse_args(argv)
+
+    return find_illegal_windows_names(args.filenames)
```
asottile, line 62: this can probably be inlined or underscored -- I don't think having a separate function improves anything here (especially once you remove the `added_files` bit)

tests/check_illegal_windows_names_test.py
```diff
@@ -0,0 +1,78 @@
...
+    find_illegal_windows_names,
+)
+from pre_commit_hooks.check_illegal_windows_names import main
+from pre_commit_hooks.check_illegal_windows_names import parents
+from pre_commit_hooks.util import cmd_output
+
+skip_win32 = pytest.mark.skipif(
+    sys.platform == 'win32',
+    reason='case conflicts between directories and files',
+)
```
asottile, line 15: I think once you remove `added_files` as a constraint you won't need to skip anything (you won't even need to have any disk state either!)

asottile (commented):
actually, crazy idea -- I think we can replace this entire hook with a `fail` hook without needing to write any code:

```yaml
-   id: check-illegal-windows-names
    name: check illegal windows names
    entry: Illegal windows filenames detected
    language: fail
    files: '(?i)(^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(\.|/|$)'
```

PR #596, opened 2021-05-11
Update flake8 redirect to new upstream home

hugovk:
Standardise https://gitlab.com/pycqa/flake8 -> https://github.com/PyCQA/flake8


asottile:
it's been a year, I'm just going to remove them

hugovk:
Okay, probably about time!

BTW I opened this after seeing the error updating a repo that was using the old hook. Seems it had only been added on 2020-12-28, maybe a copy/paste. Anyway, that one's updated now.

PR #608, opened 2021-06-02
Allow TOML 1.0.0 syntax in check-toml

hukkinj1:
`check-toml` fails with new TOML 1.0.0 syntax. This changes the TOML parser used to a spec v1.0.0 compatible one. Disclaimer: the [parser](https://github.com/hukkinj1/tomli) used is written by me.

### DIFF
--- a/pre_commit_hooks/check_toml.py
+++ b/pre_commit_hooks/check_toml.py
@@ -2,7 +2,7 @@ import argparse
 from typing import Optional
 from typing import Sequence
 
-import toml
+import tomli
 
 
 def main(argv: Optional[Sequence[str]] = None) -> int:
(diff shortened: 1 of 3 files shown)


asottile:
no thanks, `toml` is the ubiquitous library used in the python community -- surely it will be updated at some point

hukkinj1:
I respect that. Note that this is about to be [vendored in pip](https://github.com/pypa/pip/pull/10035), so pip will use it to read `pyproject.toml`.

DanielNoord:
@asottile Seeing as `tomli` is now included as `tomllib` in `3.11`, would you want a PR that switches to `tomli` similar to https://github.com/python/mypy/pull/12305 / https://github.com/pytest-dev/pytest/pull/9741?

Issue #609, opened 2021-06-03
forbid-new-submodules is broken


m-khvoinitsky:
The hook does not have an explicit `types` section.
Default `types: [file]` does not match submodules (which are directories). It works only if you commit some regular file together with committing new submodule.
Specifying `types: [directory]` helps but not much: it works as a pre-commit but doesn't as pre-push or any other way when you specify `--from-ref` and `--to-ref`.
I'd suggest stopping [ignoring provided by pre-commit list of files](https://github.com/pre-commit/pre-commit-hooks/blob/f5c42a050bf7deef0d3fa7fde327c1ae3fd4bed1/pre_commit_hooks/forbid_new_submodules.py#L8) and check `git status` for provided files instead of diff.

asottile:
@m-khvoinitsky in order to add a submodule `.gitmodules` needs to be edited -- are you sure about your assertions?  can you describe what you're seeing instead of prescribing a solution?  I suspect you've got an XY problem

m-khvoinitsky:
> in order to add a submodule .gitmodules needs to be edited

That's correct if you're adding submodules intentionally. But if user tries to commit some another git repository inside current one by accident, git complains but obediently adds a "commit" (type 160000) entry in the tree. I was hoping that this hook would prevent such situations too.

> I suspect you've got an XY problem

May be it is. If so, a separate hook which rejects type 160000 entries without corresponding records in .gitmodules suits my needs better. However, implementing proposed chages would make this hook useful in both usecases (which makes sense because rejecting "incomplete" submodule is still rejecting a submodule).
And the (separate) problem with `--from-ref` and `--to-ref` is still a problem — current hook is useless in CI.

asottile:
feel free to take a stab if you think it's worth it, I'm just not sure you're solving something that's actually a problem

PR #619, opened 2021-06-23
forbid-new-submodules fixes

m-khvoinitsky:
Fixes for #609.

### DIFF
--- a/pre_commit_hooks/forbid_new_submodules.py
+++ b/pre_commit_hooks/forbid_new_submodules.py
@@ -1,3 +1,5 @@
+import argparse
+import os
 from typing import Optional
 from typing import Sequence
 
@@ -5,10 +7,23 @@ from pre_commit_hooks.util import cmd_output
 
 
 def main(argv: Optional[Sequence[str]] = None) -> int:
-    # `argv` is ignored, pre-commit will send us a list of files that we
-    # don't care about
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*')
+    args = parser.parse_args(argv)
+
+    if (
+        'PRE_COMMIT_FROM_REF' in os.environ and
+        'PRE_COMMIT_TO_REF' in os.environ
+    ):
+        diff_arg = '...'.join((
+            os.environ['PRE_COMMIT_FROM_REF'],
+            os.environ['PRE_COMMIT_TO_REF'],
+        ))
+    else:
+        diff_arg = '--staged'
     added_diff = cmd_output(
-        'git', 'diff', '--staged', '--diff-filter=A', '--raw',
+        'git', 'diff', '--diff-filter=A', '--raw', diff_arg, '--',
+        *args.filenames,
     )
     retv = 0
     for line in added_diff.splitlines():
--- a/tests/forbid_new_submodules_test.py
+++ b/tests/forbid_new_submodules_test.py
@@ -1,22 +1,20 @@
+import os
 import subprocess
+from unittest import mock
 
 import pytest
 
 from pre_commit_hooks.forbid_new_submodules import main
+from testing.util import git_commit
 
 
 @pytest.fixture
 def git_dir_with_git_dir(tmpdir):
     with tmpdir.as_cwd():
         subprocess.check_call(('git', 'init', '.'))
-        subprocess.check_call((
-            'git', 'commit', '-m', 'init', '--allow-empty', '--no-gpg-sign',
-        ))
+        git_commit('--allow-empty', '-m', 'init')
         subprocess.check_call(('git', 'init', 'foo'))
-        subprocess.check_call(
-            ('git', 'commit', '-m', 'init', '--allow-empty', '--no-gpg-sign'),
-            cwd=str(tmpdir.join('foo')),
-        )
+        git_commit('--allow-empty', '-m', 'init', cwd=str(tmpdir.join('foo')))
         yield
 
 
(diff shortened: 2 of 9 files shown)


pre_commit_hooks/forbid_new_submodules.py
```diff
@@ -1,14 +1,26 @@
...
+import sys
 from typing import Sequence
 
 from pre_commit_hooks.util import cmd_output
 
 
-def main(argv: Optional[Sequence[str]] = None) -> int:
-    # `argv` is ignored, pre-commit will send us a list of files that we
-    # don't care about
+def main(argv: Sequence[str] = sys.argv[1:]) -> int:
```
asottile, line 8: this sets a mutable default value which does not work as you would expect in python -- please don't change this, the original argument handling was correct

```pycon
>>> def f(x=[]):
...     x.append('wat')
...     print(x)
... 
>>> f(['hi'])
['hi', 'wat']
>>> f(['hi'])
['hi', 'wat']
>>> f()
['wat']
>>> f()
['wat', 'wat']
>>> f()
['wat', 'wat', 'wat']
>>> f()
['wat', 'wat', 'wat', 'wat']
>>> f()
['wat', 'wat', 'wat', 'wat', 'wat']
```

pre_commit_hooks/forbid_new_submodules.py
```diff
@@ -1,14 +1,26 @@
...
+            '--merge-base',
+            os.environ['PRE_COMMIT_FROM_REF'],
+            os.environ['PRE_COMMIT_TO_REF'],
+        )
+    else:
+        diff_args = ('--staged',)
     added_diff = cmd_output(
-        'git', 'diff', '--staged', '--diff-filter=A', '--raw',
+        'git', 'diff', '--diff-filter=A', '--raw', *diff_args, '--', *argv,
     )
```
asottile, line 24: this shouldn't be needed -- use the arguments that are passed from pre-commit directly (you'll want to set up an `argparse.ArgumentParser` -- see how the rest of the tools do this)

tests/forbid_new_submodules_test.py
```diff
@@ -31,12 +32,34 @@ def git_dir_with_git_dir(tmpdir):
...
+    TO = subprocess.check_output(REV_PARSE_HEAD).decode().strip()
+    os.environ['PRE_COMMIT_FROM_REF'] = FROM
+    os.environ['PRE_COMMIT_TO_REF'] = TO
+    try:
+        assert main(()) == 0
+        assert main(('random_non-related_file',)) == 0
+        assert main(('foo',)) == 1
+    finally:
+        del os.environ['PRE_COMMIT_FROM_REF']
+        del os.environ['PRE_COMMIT_TO_REF']
```
asottile, line 56: use `mock.patch.dict`, though I think this code will be deleted anyway

pre_commit_hooks/forbid_new_submodules.py
m-khvoinitsky: Yep, I'm aware of the behavior of mutable defaults in python. In this case it's not a problem because the function will always be called only once and without arguments by setuptools machinery. However, I agree that using it may be fragile, for example, in tests.

pre_commit_hooks/forbid_new_submodules.py
m-khvoinitsky: Initially I thought that too and wanted to rewrite it using `git status` instead of `git diff` to query mode. However, the hook is called "forbid **new** submodules" which apparently means that bumping somehow already existent submodules should be allowed. That's why we need `--diff-filter=A` and we need to query `PRE_COMMIT_*_REF` variables in order to ensure that no submodules have been added in any commits being validated, not only the last one.

m-khvoinitsky:
Finally managed to take a look into tests failure on Windows and fix it. Please, take a look.

pre_commit_hooks/forbid_new_submodules.py
```diff
@@ -1,14 +1,32 @@
...
 
 def main(argv: Optional[Sequence[str]] = None) -> int:
-    # `argv` is ignored, pre-commit will send us a list of files that we
-    # don't care about
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*')
+    args = parser.parse_args(argv)
+
+    if not args.filenames:
+        return 0
```
asottile, line 15: pre-commit will never call this with zero arguments

pre_commit_hooks/forbid_new_submodules.py
```diff
@@ -1,14 +1,32 @@
...
+    if (
+        'PRE_COMMIT_FROM_REF' in os.environ and
+        'PRE_COMMIT_TO_REF' in os.environ
+    ):
+        diff_args: Sequence[str] = (
```
asottile, line 20: `Tuple[str, ...]`

tests/forbid_new_submodules_test.py
```diff
@@ -31,12 +33,32 @@ def git_dir_with_git_dir(tmpdir):
...
+    assert main(()) == 0
+    assert main(('random_non-related_file',)) == 0
+    assert main(('foo',)) == 1
+    out, _ = capsys.readouterr()
+    assert out.startswith('foo: new submodule introduced\n')
+
+
+def test_main_new_submodule_committed(git_dir_with_git_dir, capsys):
+    REV_PARSE_HEAD = ('git', 'rev-parse', 'HEAD')
+    FROM = subprocess.check_output(REV_PARSE_HEAD).decode().strip()
```
asottile, line 45: why are you capitalizing local variables

pre_commit_hooks/forbid_new_submodules.py
m-khvoinitsky: Yeah, I've added it because of `git diff` or `git status` behavior difference with one or more path arguments (check them) vs. zero (check _all_, not none). But I agree that this is not needed.

tests/forbid_new_submodules_test.py
m-khvoinitsky: Fixed `FROM` and `TO`, preserved `REV_PARSE_HEAD` because it's a constant. If you insist, I can put it to the top of the module (since it only needed in this function I've put it there).

asottile:
tests fail on git 2.25.1

m-khvoinitsky:
> tests fail on git 2.25.1

Fixed.

asottile (approved):

PR #630, opened 2021-07-23
Fix false positives of check-executables-have-shebangs

ylep:
`check-executables-have-shebangs` needs to check the actual file mode known to Git and avoid relying only on the filesystem permissions, because some filesystems force the executable bit on every file (e.g. fat32 on Linux).


ylep:
@asottile I am a bit disappointed to see my PR being dismissed without a word... Please let me know if I can improve it. There is a real issue with `check-executables-have-shebangs` having false positives on crippled filesystems.

asottile:
oops!  sorry I did write out a comment but I guess I clicked the wrong button

crippled filesystems are not supported -- as is clearly shown in the current code and testcases

PR #634, opened 2021-07-27
Fix inconsistent name and description values

revolter:


asottile (commented):
would actually prefer them all lower-cased as #413

revolter:
Why? 😃 They're harder to be recognized if they're lowercase.

Also, it's like using `Javascript` instead of `JavaScript`. If the [official website](https://www.json.org/) uses `JSON`, then it's `JSON`.

asottile:
why do you want them upper case?  I don't need a reason I want them lower cased

revolter:
> why do you want them upper case?

For the same reason that I want to be called `Iulian`, and not `iulian` 😂 Because that's my name.

Also, in this specific case: https://en.wikipedia.org/wiki/Acronym#All-caps_style.

PR #641, opened 2021-08-17
Fix "while scanning a plain scalar found unexpected ':'"

seroy:
I use `check-yaml` hook to validate YAML files in project.
The `pnpm` generate such sort of construction (in `pnpm-lock.yaml`):

```
github.com/account_name/package_name/md5:
  resolution: {tarball: https://codeload.github.com/account_name/package_name/tar.gz/md5}

```
And on this file the `check-yaml` raises validation error:

```
while scanning a plain scalar
  in "test.yml", line 2, column 25
found unexpected ':'
  in "test.yml", line 2, column 30
```

### DIFF
--- a/pre_commit_hooks/check_yaml.py
+++ b/pre_commit_hooks/check_yaml.py
@@ -7,7 +7,7 @@ from typing import Sequence
 
 import ruamel.yaml
 
-yaml = ruamel.yaml.YAML(typ='safe')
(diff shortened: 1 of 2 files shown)


asottile:
please create issues on the upstream project, `pure=True` is an unacceptable performance hit

seroy:
https://sourceforge.net/p/ruamel-yaml/tickets/391/ 🤦‍♂️

PR #649, opened 2021-09-14
Allow pretty-format-json module to unify and/or sort list values 

afischer-opentext-com:
This PR adds the ability to the pretty-format-json module to sort and/or unify list values, given their keys.

A use case for this are lists of resources and their permissions, which - for readability - should be sorted and unified to easily identify duplicates.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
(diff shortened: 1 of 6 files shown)


asottile:
please in the future start with an issue where new features can be discussed to save everyone time here

at an initial glance this seems far to complicated and far too niche to take on general maintenance for

afischer-opentext-com:
This is as obscure as the top-keys feature is imho. But sure, saving time, closing then.

PR #655, opened 2021-09-29
Fix failing filesystem case dependent tests on MacOS

kevoriordan:
When we try to compile pre-commit hooks with python dependencies on MacOS using https://github.com/cachix/pre-commit-hooks.nix , we get errors due to tests that depend on case sensitive filesystem failing. Adding same skip annotations for MacOS/Darwin as we have for Windows.

See here for the error output from nix build: https://gist.github.com/bbhoss/dc3d098b32395b0650e79e6d6025885c

### DIFF
--- a/tests/check_case_conflict_test.py
+++ b/tests/check_case_conflict_test.py
@@ -13,7 +13,7 @@ skip_win32 = pytest.mark.skipif(
     reason='case conflicts between directories and files',
 )
 
-skip_darwin = pytest.mark_skipif(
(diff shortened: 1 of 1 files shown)


asottile:
hi!  while I appreciate the change, we intentionally don't test against macos due to the unreliability of apple's platform so I don't want to take macos-specific patches to just the testsuite

feel free to include the patch in whatever external packaging you're planning for pre-commit-hooks however!

PR #662, opened 2021-10-04
Adjust git dir when checking for merge in worktree

### LINKED ISSUE #638: check_merge_conflict.py does not work in worktrees
https://github.com/pre-commit/pre-commit-hooks/blob/8c1183c0c8e6e1136b8ea7d47d220d0d23523476/pre_commit_hooks/check_merge_conflict.py#L17 .git is a file, not a folder in worktrees

errsyn:
Fixes #638

Right now ongoing merges aren't detected in linked working trees, as it is assumed that `.git` is a directory. However, this is not the case for linked working trees. Working trees instead have a `.git` file, which references the path to the git directory like this:

```shell
$ cat .git
gitdir: /path/to/main-working-tree/.git/worktrees/<worktree-name>
```

So when we are dealing with a linked working tree, we just have to readjust the git directory to the one referenced in the `.git` file.


pre_commit_hooks/check_merge_conflict.py
```diff
@@ -10,15 +10,25 @@
     b'=======\n',
     b'>>>>>>> ',
 ]
+GITDIR_PREFIX_LEN = len('gitdir: ')
```
asottile, line 13: I would prefer to use git's mechanisms for discovering this directory -- using implementation detail like reading the file is going to be fragile

tests/check_merge_conflict_test.py
```diff
@@ -9,21 +9,25 @@
 from testing.util import git_commit
 
 
-@pytest.fixture
-def f1_is_a_conflict_file(tmpdir):
+@pytest.fixture(scope='function', params=['repo', 'worktree'])
+def f1_is_a_conflict_file(tmpdir, request):
```
asottile, line 13: I would prefer not to use parametrized fixtures -- they're really difficult to understand and maintain

what you can actually do here instead is write a separate test which consumes this fixture and the makes a worktree from that and call the original test -- it's going to be a lot easier to follow than this action-at-a-distance

tests/check_merge_conflict_test.py
```diff
@@ -64,6 +64,22 @@ def f1_is_a_conflict_file(tmpdir):
...
+    with worktree.as_cwd():
+        cmd_output(
+            'git', 'pull', '--no-rebase', 'origin', 'master', retcode=None,
+        )
+        assert os.path.exists(
+            f1_is_a_conflict_file.join(
+                '.git', 'worktrees', 'worktree', 'MERGE_MSG',
+            ),
+        )
+        yield worktree
```
asottile, line 80: I meant that you would inline this directly in your test above the call to the other test

no need to make a fixture if only one thing is using it

tests/check_merge_conflict_test.py
```diff
@@ -98,6 +114,13 @@ def repository_pending_merge(tmpdir):
         yield repo2
 
 
+def test_not_in_a_repository(tmpdir):
+    with tmpdir.as_cwd():
+        f1 = tmpdir.join('f1')
+        f1.write('foo\n')
+        assert main(['f1']) == 0
+
+
```
asottile, line 123: I don't think this is a realistic test -- I expect these to always run inside a git repo

pre_commit_hooks/check_merge_conflict.py
```diff
@@ -13,12 +16,17 @@
 
 
 def is_in_merge() -> int:
```
asottile, line 18: this is probably my fault -- but can you change this to `-> bool`

PR #670, opened 2021-10-18
minor fix: inconsistent hook name

dennisroche:
Sorry for raising such a trival PR. 

Perhaps better solved by https://github.com/pre-commit/pre-commit-hooks/pull/634


dennisroche:
@asottile if that is your preference. IMHO that lower-case will make it harder to read.

see https://github.com/pre-commit/pre-commit-hooks/pull/671

Issue #683, opened 2021-11-05
Proposal: new hook to detect CVE-2021-42574 (Trojan Source Attacks)


yajo:
As explained in https://trojansource.codes/ and https://www.python.org/dev/peps/pep-0672/ it is very likely to receive trojan source attacks, where some UTF-8 strings contain invisible characters that make your program behave in unexpected ways.

Since those characters are proper UTF-8 characters and are invisible, even manual source code revision can bypass it. We need automated tools to do that.

It can affect many languages, so it would be great if a pre-commit hook is provided here that can forbid committing such poisoned sources.

asottile:
> it is very likely to receive trojan source attacks

I don't think that's true but it is possible

I don't think this is easy to implement especially if you are to handle homoglyphs (Α vs A) beyond banning all non-ASCII characters which isn't productive

if you can come up with a concrete proposal it may be considered but in the current state it isn't feasible

sirosen:
I was wondering about this myself. Maybe we could do this by looking at `unicodedata.category` on all characters (possibly too slow?).

I'd default to rejecting any code which contains characters in the control sequence category (`C*`), but we could also include flags to restrict it to [specific categories](https://www.unicode.org/reports/tr44/#General_Category_Values) like `Cc` and `Cf`. It's restrictive, but it's the only sensible check I can think of doing.
If this sounds valuable, I'd be happy to put together a PR.

If that's still too aggressive because it rejects legitimate control characters in strings, and we're interested in this specifically for python, I could look at writing something with `ast` or `libcst` as a separate hook.

Yajo:
Quoting from [the paper](https://trojansource.codes/trojan-source.pdf), chapter *V.F. Defenses*:

> **The simplest defense is to ban the use of text directionality
> control characters** both in language specifications and in
> compilers implementing these languages.
> 
> In most settings, this simple solution may well be suf-
> ficient. If an application wishes to print text that requires
> Bidi overrides, developers can generate those characters using
> escape sequences rather than embedding potentially dangerous
> characters into source code.

So I'd say we can just implement that simplest defense approach, at least as a start. It can serve as a shield that targets any UTF-8 compatible language.

For most projects, this should be just OK, as there is usually a workaround as explained in the quote.

This could be the next iteration, but that depends on what is a string and a comment for each language, so it becomes more complex and specific:

> This simple defense can be improved by adding a small
> amount of nuance. By banning all directionality-control char-
[...]

PR #685, opened 2021-11-11
Add check for text file encodings

scop:
Refs https://github.com/pre-commit/pre-commit-hooks/issues/683

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -32,6 +32,12 @@
     entry: check-docstring-first
     language: python
     types: [python]
+-   id: check-encoding
+    name: check text files have desired encoding
+    description: checks that text files decode cleanly using an encoding.
+    entry: check-encoding
+    language: python
+    types: [text]
 -   id: check-executables-have-shebangs
     name: check that executables have shebangs
     description: ensures that (non-binary) executables have a shebang.
--- a/README.md
+++ b/README.md
@@ -49,6 +49,11 @@ Check for files with names that would conflict on a case-insensitive filesystem 
 #### `check-docstring-first`
 Checks for a common error of placing code before the docstring.
 
+#### `check-encoding`
+Checks that text files have desired encoding.
+  - `--encoding` - specify encoding to assert; if not specified, default is
+    [platform dependent](https://docs.python.org/3/library/functions.html?highlight=encoding#open)
+
 #### `check-executables-have-shebangs`
 Checks that non-binary executables have a proper shebang.
 
--- a/pre_commit_hooks/check_encoding.py
+++ b/pre_commit_hooks/check_encoding.py
@@ -1,0 +1,29 @@
+import argparse
+from typing import Optional
+from typing import Sequence
+
+
+def main(argv: Optional[Sequence[str]] = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument('filenames', nargs='*', help='Filenames to check.')
+    parser.add_argument('--encoding', help='Encoding to assert.')
+    args = parser.parse_args(argv)
+
+    retval = 0
+    for filename in args.filenames:
+        try:
+            with open(filename, encoding=args.encoding) as f:
+                f.read()
+        except LookupError as exc:
+            # Unknown encoding, don't bother with the rest
+            print(f'{__file__}: {exc}')
+            retval = 2
+            break
+        except Exception as exc:
+            print(f'{filename}: {exc}')
+            retval = 1
+    return retval
+
+
+if __name__ == '__main__':
+    raise SystemExit(main())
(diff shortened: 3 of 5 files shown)


scop:
Don't know if I'm too happy with the platform dependent default, perhaps defaulting to utf-8 would be better. No strong opinions between those.

scop:
Also, incremental decoding instead of `read()` slurp would be nicer.

Can polish once we establish this will be accepted in the first place.

.pre-commit-hooks.yaml
```diff
@@ -32,6 +32,12 @@
     entry: check-docstring-first
     language: python
     types: [python]
+-   id: check-encoding
+    name: check text files have desired encoding
+    description: checks that text files decode cleanly using an encoding.
+    entry: check-encoding
+    language: python
```
Yajo, line 39: Why only python? 🤔

scop: `language` means what the hook is written in. One line below is what the hook applies _to_:
```
types: [text]
```

Yajo: 🤦🏼‍♂️ sorry

scop:
From https://github.com/pre-commit/pre-commit-hooks/issues/683#issuecomment-970279641
> any decent programming language already presents encoding issues as syntax errors so I'm not sure where it would be useful

Sadly, not all developers run test suites or test their code before pushing code to repos. Not all repos even have test suites for that matter. It's useful to be able to catch errors locally in pre-commit before they hit CI, or even staging/production.

Anyway, I think the opinion what qualifies as a "decent programming language" belongs here.

No matter what one thinks of them, for example C and JavaScript and Perl code can be encoded let's say in ISO-8859-1 without any encoding markers. I don't think there even _is_ a way to convey that at least for C code, not sure for JS; for Perl there is `use utf8;` but I don't know if something similar exists for other encodings, and there might be reasons why one wants to have files in some particular, other encoding. And such files compile and run just fine.

Chances are that the ISO-8859-1 in them is not the intention though and they now don't behave correctly, perhaps one wanted UTF-8 instead. Or perhaps one wanted code that is not that brittle wrt source encoding in the first place and for that reason wants to restrict everything to ASCII to avoid bugs of this class altogether.

Or maybe it's not a runtime issue, one just wanted some comments/API docs in them in UTF-8 but a badly configured editor did it in ISO-8859-1 or vice versa, and now docs are not being generated right. This check would help.

From https://github.com/pre-commit/pre-commit-hooks/issues/683#issuecomment-970334744 
> I understand but the vast majority of programming is writing code so I don't think it's that useful

What I'm trying to say is that just like pre-commit, this check can be useful for repos that don't have anything whatsoever to do with programming. Websites built with markdown, static HTML + CSS, other text content, including plain text, for whatever purpose.

asottile:
> > any decent programming language already presents encoding issues as syntax errors so I'm not sure where it would be useful
>
> Sadly, not all developers run test suites or test their code before pushing code to repos. Not all repos even have test suites for that matter. It's useful to be able to catch errors locally in pre-commit before they hit CI, or even staging/production.

yeah agreed, but I don't think it's our place to supply those checks when a language-specific tool can provide better diagnostic SyntaxErrors when it's a problem for them

___

my main concerns with this are:

- it's not something I would ever see myself using, or recommending to others
- it should default to UTF-8 and not platform specific
- it's being proposed as a solution to bidi / homoglyphs / etc. of which I strongly believe it is not a proper solution to this
- I think this particular "use X encoding for Y language" is better solved in language-specific tooling
- if we accept this it'll be something more to maintain, I can already see the "please read my emacs comment headers and use that encoding" request which I really don't have an interest in supporting

scop:
Some counterarguments:
- No problem changing the default, UTF-8 probably is the better choice.
- We can forget about the #683 considerations, and people who know what they're doing can do what they want with this.
- I don't think language specific tooling is really possible for a variety of cases this applies to.
- There already is a bunch of checks here that hardcode an encoding or do the right thing with some specific ones only, and there's no support for those comments headers or editorconfigs or whatnot in them either.

tspearconquest:
This would be useful in checking any kind of text files are encoded in the desired user-configurable encoding. Regardless of whether it treats or mitigates the CVE, this has value in tons of repositories.

Also, language-specific tooling means a lot of different things to a lot of people. I use vim and write a lot of terraform code and yaml for kubernetes deployments.

At least for Terraform, the cli will error out if, for example, you read a file with the `file()` function and that file "contains invalid UTF-8 sequences." However, that's not really what I'd call "language-specific tooling" because it depends on the user actually trying to run `terraform apply` whereas what I need is exactly this pre-commit hook to prevent files that have any encoding other than UTF-8 from being committed in the first place.

Respectfully, I request the addition of this hook.

asottile:
appreciate the PR but I think ultimately I don't want to take on this maintenance

PR #689, opened 2021-11-19
Suggest usable call when executable files lack shebang on Windows

### LINKED ISSUE #686: Don't suggest chmod +/-x on Windows
Some of the executable error messages suggest using `chmod +/-x` to toggle the executable bits. This works fine on *nix, but not on windows, and the results range from 'command not found' to doing nothing at all.

Eg: https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/check_executables_have_shebangs.py#L66

Instead, this should recommend `git add --chmod=+/-x somepath` when the hook detects it's running on windows.

pujitm:
Resolves #686 

> Some of the executable error messages suggest using chmod +/-x to toggle the executable bits. This works fine on *nix, but not on windows, and the results range from 'command not found' to doing nothing at all.
> 
> Eg: https://github.com/pre-commit/pre-commit-hooks/blob/56b4a7e506901ff86f8de5c2551bc41f8eacf717/pre_commit_hooks/check_executables_have_shebangs.py#L66
> 
> Instead, this should recommend `git add --chmod=+/-x somepath` when the hook detects it's running on windows.

### DIFF
--- a/pre_commit_hooks/check_executables_have_shebangs.py
+++ b/pre_commit_hooks/check_executables_have_shebangs.py
@@ -64,6 +64,8 @@ def _message(path: str) -> None:
         f'{path}: marked executable but has no (or invalid) shebang!\n'
         f"  If it isn't supposed to be executable, try: "
         f'`chmod -x {shlex.quote(path)}`\n'
+        f'  If on Windows, you may also need to: '
+        f'`git add --chmod=-x {shlex.quote(path)}`\n'
         f'  If it is supposed to be executable, double-check its shebang.',
         file=sys.stderr,
     )


pujitm:
I also noticed that platform checks occur in other functions, like the `check_executables` function:
https://github.com/pre-commit/pre-commit-hooks/blob/56b4a7e506901ff86f8de5c2551bc41f8eacf717/pre_commit_hooks/check_executables_have_shebangs.py#L19

It might be helpful to refactor those into a consistent call for platform/os checks.

pre_commit_hooks/check_executables_have_shebangs.py
```diff
@@ -60,10 +60,12 @@ def has_shebang(path: str) -> int:
 
 
 def _message(path: str) -> None:
+    is_windows = sys.platform == 'win32'
+    remove_execute_permissions_command = 'git add --chmod=-x' if is_windows else 'chmod -x'
```
asottile, line 64: I would just write this literally in the messaging rather

Kurt-von-Laven:
Happy to file a separate issue for this if it is out of scope for this pull request, but running `check-executables-have-shebangs` from WSL (e.g., Ubuntu) on the Windows file system results in "marked executable but has no (or invalid) shebang!" for every file without a shebang. In this (and every other) situation, one could instead parse the output of [git ls-tree](https://git-scm.com/docs/git-ls-tree) to determine which files are executable. Another option would be to run `git add --chmod=-x` on all files without a shebang, which should work on both Windows and Linux, but would change the behavior of the hook from read-only to auto-fix, and has the drawback of being anti-helpful when the file was supposed to be marked executable but was lacking a shebang.

asottile:
> Happy to file a separate issue for this if it is out of scope for this pull request, but running `check-executables-have-shebangs` from WSL (e.g., Ubuntu) on the Windows file system results in "marked executable but has no (or invalid) shebang!" for every file without a shebang. In this (and every other) situation, one could instead parse the output of [git ls-tree](https://git-scm.com/docs/git-ls-tree) to determine which files are executable. Another option would be to run `git add --chmod=-x` on all files without a shebang, which should work on both Windows and Linux, but would change the behavior of the hook from read-only to auto-fix, and has the drawback of being anti-helpful when the file was supposed to be marked executable but was lacking a shebang.

unrelated, and arguably the tool is correct in that case (the files from windows have `+x` on them inside WSL)

pre_commit_hooks/check_executables_have_shebangs.py
```diff
@@ -60,10 +60,11 @@ def has_shebang(path: str) -> int:
 
 
 def _message(path: str) -> None:
+    is_windows = sys.platform == 'win32'
     print(
         f'{path}: marked executable but has no (or invalid) shebang!\n'
         f"  If it isn't supposed to be executable, try: "
-        f'`chmod -x {shlex.quote(path)}`\n'
+        f'`{"git add --chmod=-x" if is_windows else "chmod -x"} {shlex.quote(path)}`\n'
```
asottile, line 67: ah I mean, without having any branching logic just write:

"chmod -x ..."
" if on windows you may also need to ..."

asottile (approved):

PR #697, opened 2021-12-01
Skip case insensitivity tests on darwin.

fiadliel:
MacOS usually uses a case-insensitive filesystem, and that causes these tests to fail with `FileExistsError`.


asottile:
dupe #655

fiadliel:
Sorry, I looked in issues, but not PRs.

PR #713, opened 2022-01-06
update display name of `name-tests-test`

asottile:


klaerik:
Thanks for this update. Was just here looking at how to do it myself. 

This should make it less confusing for groups that use the `test_.*py` convention since it shows up correctly after the hooks run.

PR #714, opened 2022-01-07
Fix misleading error message

roskakori:
The command given in the previous error message only changed the local script to executable. When running pre-commit the next time, it failed again with the same error message.

The new command actually changes the script in the repository to be executable, and pre-commit runs through the next time.

See also: [Make Shell Script Committed to Git Executable](https://www.pixelninja.me/make-script-committed-to-git-executable/).

### DIFF
--- a/pre_commit_hooks/check_shebang_scripts_are_executable.py
+++ b/pre_commit_hooks/check_shebang_scripts_are_executable.py
@@ -34,7 +34,7 @@ def _message(path: str) -> None:
     print(
         f'{path}: has a shebang but is not marked executable!\n'
         f'  If it is supposed to be executable, try: '
-        f'`chmod +x {shlex.quote(path)}`\n'
+        f'`git update-index --chmod=+x {shlex.quote(path)}`\n'
         f'  If it not supposed to be executable, double-check its shebang '
         f'is wanted.\n',
         file=sys.stderr,


asottile:
can you show console output? it works correctly for me:

```console
$ tail -n999 t.sh .pre-commit-config.yaml 
==> t.sh <==
#!/usr/bin/env bash

==> .pre-commit-config.yaml <==
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.0.1
    hooks:
    -   id: check-shebang-scripts-are-executable
$ git commit -m "add t.sh"
Check that scripts with shebangs are executable..........................Failed
- hook id: check-shebang-scripts-are-executable
- exit code: 1

t.sh: has a shebang but is not marked executable!
  If it is supposed to be executable, try: `chmod +x t.sh`
  If it not supposed to be executable, double-check its shebang is wanted.

$ chmod +x t.sh
$ git add -u
$ git commit -m "add t.sh"
Check that scripts with shebangs are executable..........................Passed
[master (root-commit) 99b3b48] add t.sh
 2 files changed, 6 insertions(+)
 create mode 100644 .pre-commit-config.yaml
 create mode 100755 t.sh
```

roskakori:
@asottile The commands you were using:
```bash
chmod +x t.sh
git add -u
```
However, the original error message only tells about the
```
chmod +x t.sh
```
If you do only that, git does not know about the change and it also wont show in IDE's as commitable.

The combination of `chmod` and `git add` works, but why issue two commands if one can do it?

asottile:
1. git absolutely does know, if your IDE doesn't that's their bug:

```console
$ git status
On branch master
nothing to commit, working tree clean
$ chmod +x t.sh 
$ git status
On branch master
Changes not staged for commit:
	modified:   t.sh

no changes added to commit
```

2. your command doesn't work

```diff
$ git update-index --chmod=+x t.sh
$ ./t.sh
bash: ./t.sh: Permission denied
```

roskakori:
> yeah the suggested patch here is not correct, it does not make the file executable

Hm, I think I get it now. You actually need to commands. One to make it locally executable, and another to update it in git.

I did not realize that because I first tried the "chmod" and when that did not work added the "git update-index", which in combination made it work.

So the correct error message would be:
> If it is supposed to be executable, try:  
>   chmod +x t.sh
>   git add --update t.sh

asottile:
yeah except there's no need, adding to git is assumed.  flake8 doesn't tell you to fix unused variables and then run `git add`

roskakori:
> yeah except there's no need, adding to git is assumed. flake8 doesn't tell you to fix unused variables and then run `git add`

After a code fix the source automatically shows on the "uncommited changes". After a chmod is does not. That what makes it confusing.

asottile:
it does though, as above:

> ```
> $ git status
> On branch master
> nothing to commit, working tree clean
> $ chmod +x t.sh 
> $ git status
> On branch master
> Changes not staged for commit:
> 	modified:   t.sh
> 
> no changes added to commit
> ```

plus you'd have to ignore this bright yellow warning:

image

roskakori:
> it does though, as above:

Oops, just figured that myself. I was using PyCharm and it seems to have a delay with the automatic update of the commit view.

My bad, sorry for wasting your time.

PR #730, opened 2022-02-19
Check Git core.fileMode rather than infer from OS.

### LINKED ISSUE #690: `check-executables-have-shebangs` Should Check Executable Bit According to Git Rather Than File System
Running `check-executables-have-shebangs` from WSL (e.g., Ubuntu) on the Windows file system results in "marked executable but has no (or invalid) shebang!" for every file without a shebang. In this (and every other) situation, one could instead parse the output of `git ls-tree` to determine which files are executable. Another option would be to run `git add --chmod=-x` on all files without a shebang, which should work on both Windows and Linux, but would change the behavior of the hook from read-only to auto-fix, and has the drawback of being anti-helpful when the file was supposed to be marked executable but was lacking a shebang.

Kurt-von-Laven:
There was already a guard preventing the `check-executables-have-shebangs` hook from raising false positives on win32 by looking up the Git file mode rather than relying on the file mode in the file system. Git already automatically probes the file system for executable bit support. Leverage Git's [`core.fileMode` config variable](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode) to prevent false positives on all file systems that don't track executable bits.

Fixes #690.

### DIFF
--- a/pre_commit_hooks/check_executables_have_shebangs.py
+++ b/pre_commit_hooks/check_executables_have_shebangs.py
@@ -15,7 +15,10 @@ EXECUTABLE_VALUES = frozenset(('1', '3', '5', '7'))
 
 
 def check_executables(paths: list[str]) -> int:
-    if sys.platform == 'win32':  # pragma: win32 cover
+    fs_tracks_executable_bit = cmd_output(
+        'git', 'config', 'core.fileMode', retcode=None,
+    ).strip()
+    if fs_tracks_executable_bit == 'false':  # pragma: win32 cover
         return _check_git_filemode(paths)
     else:  # pragma: win32 no cover
         retv = 0


pre_commit_hooks/check_executables_have_shebangs.py
```diff
@@ -8,23 +8,36 @@
...
-        for path in paths:
-            if not has_shebang(path):
-                _message(path)
-                retv = 1
-
-        return retv
+    fs_tracks_executable_bit = cmd_output(
+        'git', 'config', 'core.fileMode', retcode=None,
+    )
+    if fs_tracks_executable_bit != 'true':  # pragma: win32 cover
```
asottile, line 22: this condition is always false so your entire testsuite and changes in this and the other PR are incorrect

pre_commit_hooks/check_executables_have_shebangs.py
```diff
@@ -8,23 +8,36 @@
...
 
+from pre_commit_hooks.util import CalledProcessError
 from pre_commit_hooks.util import cmd_output
 from pre_commit_hooks.util import zsplit
 
 EXECUTABLE_VALUES = frozenset(('1', '3', '5', '7'))
 
 
 def check_executables(paths: list[str]) -> int:
-    if sys.platform == 'win32':  # pragma: win32 cover
```
asottile, line 18: feels to me like this should be a 2-line patch here instead of rewriting half the file

Kurt-von-Laven (commented):
Thanks for spotting my trailing newline (`'true\n' != 'true'`) bug.

pre_commit_hooks/check_executables_have_shebangs.py
asottile: the coverage issues you experienced _were caused by your mistake_

Kurt-von-Laven: I don't see a connection and just pushed a test commit on both PRs demonstrating that all of the pragma statements remain necessary in the current state of both branches. Again, if I'm missing something, please let me know, and I will gladly correct the mistake. I will, of course, take the test commits back down after you have a chance to respond.

asottile: I don't really get what you don't understand -- if you make the two line patch and leave the current pragmas in place you'll pass coverage

asottile: and if it doesn't pass coverage, that shows that your change is incorrect

Kurt-von-Laven: The patch I started from (`!= 'true'`) was not the same one you had in mind (the simpler `== 'false'`). Again though, the pragmas were correct at every iteration, and my initial bug was not merely a result of disabling a check that should have been left in place. The previous iteration was functionally correct if verbose.

Kurt-von-Laven: To clarify, I simplified it to use `== 'false'` instead.

Kurt-von-Laven:
I believe this pull request is ready to be merged.

asottile (approved):

PR #734, opened 2022-03-01
Add check for permitted characters in pathnames

noahp:
Add a new checker that checks for allowed characters in pathnames.

Includes arguments for overriding the default allowed list and extending
it.

Default allow list is alphanumeric plus `.-_`.

### DIFF
--- a/pre_commit_hooks/check_permitted_path_characters.py
+++ b/pre_commit_hooks/check_permitted_path_characters.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+import argparse
+import string
+from typing import Sequence
+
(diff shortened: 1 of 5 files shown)


pre_commit_hooks/check_permitted_path_characters.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+import argparse
+import string
+from typing import Sequence
+
+DEFAULT_ALLOWLIST = string.ascii_letters + string.digits + '-_./'
```
noahp, line 7: `os.path.sep` or are paths unix-normalized by the time they arrive at the hook scripts?

asottile:
please start new feature development with an issue first that way we can discuss the idea before committing time to it

PR #743, opened 2022-03-23
Default allow_missing_credentials flag to True

droctothorpe:
I just had to help an engineer troubleshoot this failing pre-commit hook because he had never used the AWS CLI / SDK and didn't have any AWS credentials on his local machines. IMHO, the `allow_missing_credentials` flag should default to `True`. It's hard to imagine a use case in which you would want to force someone to have AWS credentials on their local machine in order to make sure that they didn't put said AWS credentials (that are _not_ on their local machine 🙃 ) in the specified codebase.

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -107,6 +107,7 @@ def main(argv: Sequence[str] | None = None) -> int:
     parser.add_argument(
         '--allow-missing-credentials',
         dest='allow_missing_credentials',
+        default=True,
         action='store_true',
         help='Allow hook to pass when no credentials are detected.',
     )


asottile:
this is a breaking change and a reduction in "security" -- I think the hook shouldn't blindly default to allowing creds if it doesn't identify yours

so unfortunately I can't accept this patch -- I'd recommend setting the `args` in your configuration to the value if that's what you want it to do

droctothorpe:
Yeah, that's precisely what we did. I don't think defaulting to `True` is equivalent to allowing creds. It's more like not looking for a bounty (key) when there's nothing posted on the bounty board (aws config / env vars), but yeah no sweat, we have a functional solution on our end. Thanks for responding so quickly. I can see why the star to open PR ratio is so good for this repo! 🙃

PR #746, opened 2022-04-01
added dbgp (Komodo IDE debugger) package

agroszer:


agroszer:
ohwell, https://pypi.org/project/komodo-python3-dbgp/ is just a tad better and Komodo isn't very widely used either

asottile:
that seems dead as well -- and not popular enough to make the list?

agroszer:
where is the bar to make the list?

asottile:
there isn't an explicit one -- but a <5-star library that hasn't been updated in a few years doesn't seem active enough

additionally the need for such a list is diminishing due to `breakpoint`

PR #750, opened 2022-04-10
Support checking executable bit without Git.

### LINKED ISSUE #749: `check-shebang-scripts-are-executable` Assumes Git Repo
`check-executables-have-shebangs` supports being run on files outside of a Git repository, so it seems natural for `check-shebang-scripts-are-executable` to do the same. I would guess it's also slightly faster to check executable bits through the file system when it can be trusted to track them properly.

Kurt-von-Laven:
The `check-shebang-scripts-are-executable` hook already avoided false negatives in a Git repository by looking up the Git file mode rather than relying on the file mode in the file system. Git already automatically probes the file system for executable bit support. Use the file mode in the file system when we are not in a Git clone or it is trusted by Git according to its `core.fileMode` config variable.

Also, fix previously asymptomatic copy/paste bug in `check_shebang_scripts_are_executable_test.test_git_executable_shebang` exposed by the other commit. This test manually filtered executable files out before calling `check_shebang_scripts_are_executable`. This makes sense in the test it was copied from, `check_executables_have_shebangs.test_git_executable_shebang`, because the `check-executables-have-shebangs` hook only runs on executable files. However, `check-shebang-scripts-are-executable` correctly runs on all text files, so the test shouldn't filter executable files out. The test still passed because when `git ls-files` is passed no files in particular, it
lists all files in the Git repository that satisfy the given filters.

Closes #749.

### DIFF
--- a/pre_commit_hooks/check_shebang_scripts_are_executable.py
+++ b/pre_commit_hooks/check_shebang_scripts_are_executable.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import argparse
+import os
 import shlex
 import sys
 from typing import Sequence
@@ -9,13 +10,30 @@ from typing import Sequence
 from pre_commit_hooks.check_executables_have_shebangs import EXECUTABLE_VALUES
 from pre_commit_hooks.check_executables_have_shebangs import git_ls_files
 from pre_commit_hooks.check_executables_have_shebangs import has_shebang
+from pre_commit_hooks.util import cmd_output
 
 
 def check_shebangs(paths: list[str]) -> int:
-    # Cannot optimize on non-executability here if we intend this check to
-    # work on win32 -- and that's where problems caused by non-executability
-    # (elsewhere) are most likely to arise from.
-    return _check_git_filemode(paths)
+    fs_tracks_executable_bit = cmd_output(
+        'git', 'config', 'core.fileMode', retcode=None,
+    ).strip()
+    return (
+        _check_git_filemode(paths)
+        if fs_tracks_executable_bit == 'false'
+        else _check_fs_filemode(paths)
+    )
+
+
+def _check_fs_filemode(
+    paths: list[str],
+) -> int:  # pragma: win32 no cover
+    retv = 0
+    for path in paths:
+        if not os.access(path, os.X_OK) and has_shebang(path):
+            _message(path)
+            retv = 1
+
+    return retv
 
 
 def _check_git_filemode(paths: Sequence[str]) -> int:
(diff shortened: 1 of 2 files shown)


tests/check_shebang_scripts_are_executable_test.py
```diff
@@ -9,6 +9,62 @@
 from pre_commit_hooks.check_shebang_scripts_are_executable import main
 from pre_commit_hooks.util import cmd_output
 
+skip_win32 = pytest.mark.skipif(
```
asottile, line 12: these should be xfail if anything -- then you shouldn't need the no cover

tests/check_shebang_scripts_are_executable_test.py
```diff
@@ -83,7 +139,5 @@ def test_git_executable_shebang(temp_git_dir, content, mode, expected):
         cmd_output('chmod', mode, str(path))
         cmd_output('git', 'update-index', f'--chmod={mode}', str(path))
 
-        # simulate how identify chooses that something is executable
-        filenames = [path for path in [str(path)] if os.access(path, os.X_OK)]
```
asottile, line 87: this bit was important to demonstrate the windows behaviour

pre_commit_hooks/check_shebang_scripts_are_executable.py
```diff
@@ -2,20 +2,36 @@
...
-    # Cannot optimize on non-executability here if we intend this check to
-    # work on win32 -- and that's where problems caused by non-executability
-    # (elsewhere) are most likely to arise from.
-    return _check_git_filemode(paths)
+    fs_tracks_executable_bit = cmd_output(
+        'git', 'config', 'core.fileMode', retcode=None,
+    ).strip()
+    return _check_git_filemode(paths) \
+        if fs_tracks_executable_bit == 'false' \
+        else _check_fs_filemode(paths)
```
asottile, line 22: don't use backslashes please

tests/check_shebang_scripts_are_executable_test.py
Kurt-von-Laven: Neat function. I tried removing the no cover and got [a test coverage failure](https://asottile.visualstudio.com/asottile/_build/results?buildId=12893&view=logs&j=c5301fb6-0b07-5e6e-20f2-e78065fa524c&t=60800355-5684-5485-0283-7dd0b5fe8f8f) though. Did I miss something here?

Kurt-von-Laven:
@asotille, please let me know if there's anything I should change about this pull request.

PR #766, opened 2022-05-11
Update eof fixer

jf:
This updates end_of_file_fixer.py. The commit messages should be self-explanatory, but here are a few notes:

For https://github.com/pre-commit/pre-commit-hooks/commit/67b1980f3e7396223928cd9a1dd7469f5915f28d, unless the comment on lines 10-11 are wrong, and I missed something in testing, there is no need to test for an empty file. https://github.com/pre-commit/pre-commit-hooks/blob/8f6152921e65fe1a82bc475cafb721bd525ba5df/pre_commit_hooks/end_of_file_fixer.py#L14 (`except OSError`) already takes care of empty files

For https://github.com/pre-commit/pre-commit-hooks/commit/36df47b046eee6af0816d6f366eaa5ef718006e6, I thought it made for sense to say "Fixed", rather than "Fixing", given that the file is already fixed by the time we print the message

### DIFF
--- a/pre_commit_hooks/end_of_file_fixer.py
+++ b/pre_commit_hooks/end_of_file_fixer.py
@@ -62,7 +62,7 @@ def main(argv: Sequence[str] | None = None) -> int:
         with open(filename, 'rb+') as file_obj:
             ret_for_file = fix_file(file_obj)
             if ret_for_file:
-                print(f'Fixing {filename}')
+                print(f'Fixed {filename}')
             retv |= ret_for_file
 
     return retv
--- a/pre_commit_hooks/end_of_file_fixer.py
+++ b/pre_commit_hooks/end_of_file_fixer.py
@@ -13,9 +13,10 @@ def fix_file(file_obj: IO[bytes]) -> int:
         file_obj.seek(-1, os.SEEK_END)
     except OSError:
         return 0
+
     last_character = file_obj.read(1)
-    # last_character will be '' for an empty file
-    if last_character not in {b'\n', b'\r'} and last_character != b'':
+
+    if last_character not in {b'\n', b'\r'}:
         # Needs this seek for windows, otherwise IOError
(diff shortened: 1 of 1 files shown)


jf:
re https://asottile.visualstudio.com/751f6930-3111-476a-9fff-020f43cda3da/_build/results?buildId=12854: hm, it looks like IRL with real files we don't need `last_character != b''`, but it is there to make the tests pass?

asottile:
your assertion about empty files is wrong as the tests show

asottile:
are you seeing actual breakage?  or does this patch improve nothing

jf:
is there a use case where the hook needs to be called on a string instead of a file? I'm not seeing `file_obj.seek(-1, os.SEEK_EEND)` raise OSError when `file_obj` is based on `io.BytesIO` instead of an actual file. The way it looks to me, `test_fix_file` is basically the same as `test_integration` in `tests/end_of_file_fixer_test.py`, and the code would not break `test_integration`

jf:
> are you seeing actual breakage? or does this patch improve nothing

it makes the code more readable IMO. We wouldn't have to keep extra code `and last_character != b''` just to make a test pass that's based on passing strings instead of files

asottile:
ok, well then please in the future don't send patches unless there's an actual improvement

my time is already limited and if I'm spending it on inconsequential cosmetic changes (that might even be regressions!) then everything is worse

this isn't specific to me, this is just how to be a good open source citizen

jf:
> ok, well then please in the future don't send patches unless there's an actual improvement
> 
> my time is already limited and if I'm spending it on inconsequential cosmetic changes (that might even be regressions!) then everything is worse
> 
> this isn't specific to me, this is just how to be a good open source citizen

sorry I just thought this was an improvement! I guess we have different definitions here of what means. I just like readable code that's easier to understand. Sorry for wasting your time

PR #768, opened 2022-05-11
improve description for end-of-file-fixer

jf:


asottile:
I'm not sure this changes anything -- was there confusion about what "one newline" means?

jf:
technically, a line that has 4 newlines at the bottom also ends with one newline? just wanting to be more precise, especially since I do see in the code that extra newlines are removed.

asottile:
> technically, a line that has 4 newlines at the bottom also ends with one newline? just wanting to be more precise, especially since I do see in the code that extra newlines are removed.

no it doesn't it ends with 4

jf:
ok

jf:
sorry just one last thing because I just happened to see it: https://github.com/pre-commit/pre-commit-hooks/blob/aa56da853574fa6128b06cf53f5b6bdbc16d4534/README.md#end-of-file-fixer: `Makes sure files end in a newline and only a newline.` Just FYI.

Not absolutely crucial, but probably good to harmonize the descriptions as much as possible between `README.md` and `.pre-commit-hooks.yaml`

PR #771, opened 2022-05-24
Use tomli and tomllib instead of toml

### LINKED ISSUE #755: Support newer TOML syntax
With the recent adoption of `tomli` as PEP680 `tomllib`, the `toml` package (which supports only TOML v0.5.0) used to verify syntactic correctness of toml files should be seen as deprecated. In #608 a switch to `tomli` was rejected, however given that the situation for `toml` has not improved since then, the `pre-commit-hooks` should switch as well. Alternatively the project could switch to `rtoml` which also supports newer versions of the TOML standard.

DanielNoord:
Closes #755.

Originally this switch was rejected because the hope was that `toml` would be updated at one point, see https://github.com/pre-commit/pre-commit-hooks/pull/608. However, the latest commit is from 1 November 2020 ([commits](https://github.com/uiri/toml/commits/master)) and `tomli` will be part of Python `3.11` so it seems like a good candidate to replace it.

Similar PRs have been used in other "mainstream" projects like https://github.com/python/mypy/pull/12305 and https://github.com/pytest-dev/pytest/pull/9741.

### DIFF
--- a/pre_commit_hooks/check_toml.py
+++ b/pre_commit_hooks/check_toml.py
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
 import argparse
+import sys
 from typing import Sequence
 
-import toml
+if sys.version_info >= (3, 11):  # pragma: >=3.11 cover
+    import tomllib
+else:  # pragma: <3.11 cover
+    import tomli as tomllib
 
 
 def main(argv: Sequence[str] | None = None) -> int:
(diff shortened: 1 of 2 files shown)


DanielNoord:
CI seems to fail because of lack of coverage on L8. I could add a `pragma: no cover` comment if desired.

asottile:
the correct pragmas would be `# pragma: >=3.11 cover` and `# pragma: <3.11 cover`

DanielNoord:
> the correct pragmas would be `# pragma: >=3.11 cover` and `# pragma: <3.11 cover`

Added! 😄

asottile (approved):

PR #782, opened 2022-06-10
Add parameter controlling whether lfs files are excluded from rule

adam-zethraeus:
Add `—allow-in-lfs=$BOOL` parameter to `check_added_large_files` to enable preventing large files from being stored — even if git lfs is available.

The parameter defaults to `true`. This maintains the current behavior of permitting large files as long as they are in lfs.

### DIFF
--- a/pre_commit_hooks/check_added_large_files.py
+++ b/pre_commit_hooks/check_added_large_files.py
@@ -41,7 +41,7 @@ def find_large_added_files(
     # us about
     retv = 0
     filenames_filtered = set(filenames)
-    
+
     if allow_in_lfs:
         filter_lfs_files(filenames_filtered)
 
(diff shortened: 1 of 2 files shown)


asottile:
in the future please start with an issue describing your feature and rationale before spending time on it

I don't think this really makes sense as part of this tool, if you want to forbid lfs that's a separate topic entirely from forbidding large files

unrelated because I don't think I'm accepting this, you want `action='store_true'` and not `type=bool` which will just always be truthy

```pycon
>>> import argparse
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--wat', type=bool)
_StoreAction(option_strings=['--wat'], dest='wat', nargs=None, const=None, default=None, type=<class 'bool'>, choices=None, help=None, metavar=None)
>>> parser.parse_args(('--wat=true',))
Namespace(wat=True)
>>> parser.parse_args(('--wat=false',))
Namespace(wat=True)
```

your test also doesn't demonstrate your behaviour because of this

adam-zethraeus:
Eep. Sorry for the likely notification churn. Hadn’t expected anyone to hop on this. And I appreciate the tips :)

My goal here was to forbid large files — and *not* allow the git lfs carveout that the current implementation accepts.
I’ll open a proper issue to discuss tomorrow.

PR #805, opened 2022-09-14
`name-tests-test`: Search tests in both `tests` and `test`

hofbi:
This makes the `s` in `tests` optional to allow both `tests` and `test` for placing the unit tests


asottile:
the point is to enforce a standard, if you want to deviate from that you can override `files` in your repository

hofbi:
Just out of curiosity, is there an official recommendation (PEP or similar) that suggests to use `tests` over `test`. I tired to find it before opening the PR but could not find it.

PR #812, opened 2022-10-03
Fix 159: check-docstring-first shall ignore attribute docstring at module top level

joaoe:
Now attribute docstrings no longer complain.
https://peps.python.org/pep-0257/#what-is-a-docstring

### DIFF
--- a/pre_commit_hooks/check_docstring_first.py
+++ b/pre_commit_hooks/check_docstring_first.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import ast
 import io
 import tokenize
 from tokenize import tokenize as tokenize_tokenize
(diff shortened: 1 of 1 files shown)


asottile:
this breaks the point of the tool entirely -- if it isn't something you want to enforce I'd recommend just removing it instead

asottile:
#159 is closed because this is "unfixable" but for some who do not (ab)use inline docstrings this is still valuable

joaoe:
@asottile 

> if it isn't something you want to enforce I'd recommend just removing it instead

Well, there is plenty of value of putting docstrings before imports, but the tool could ignore further docstrings further down the file.

How about protecting the behavior with a command line switch ? 

> for some who do not (ab)use inline docstrings this is still valuable

It is not (ab)use, it is following PEP 257 which this tool violates.

asottile:
no thanks

PR #817, opened 2022-10-06
Double quote to single quote in Jupyter notebooks

cyyc1:
I'm extending the double quote fixer to also cover Jupyter notebooks (.ipynb) files, because more and more people write Python code in .ipynb so it would be nice if pre-commit can provide support.


asottile:
hi! looks good but we're not accepting new language specific hooks here -- (I'm even planning to take the current ones and move them out at some point)

in the future please open an issue and discuss it first before working on it (just general open source advice)

asottile:
> > hi! looks good but we're not accepting new language specific hooks here -- (I'm even planning to take the current ones and move them out at some point)
> > in the future please open an issue and discuss it first before working on it (just general open source advice)
> 
> Hi @asottile thanks for the reply.
> 
> I plan to make this a standalone pre-commit hook then. Since I plan to reuse a part of the code in string_fixer.py, I suppose I need to include proper citation or copyright notice as per your MIT license.
> 
> Do you have some suggestions on the format for citation? You are right that I'm new to writing open source code. Any suggestion is appreciated!

all that the license requires is a copy of the license -- I usually put derivative licenses in a `./licenses` directory and include them with `license_files` in setup.cfg / setup.py

cyyc1:
Thanks. Good to know!
Btw your videos on how to write a flake8 plugin and autoformatter were quite helpful. They inspired me to write 3 small tools.

Issue #780, opened 2022-06-07
`pretty-format-json` modifies floating point numbers that have too many digits of precision


dougthor42:
`pretty-format-json` modifies floating point numbers that have too many digits of precision.

### Background

I have to work with JSON files that may be generated by non-python programs. These files, for whatever reason, have numbers that have up to 16 digits after the decimal place.

(It's absurd, really. The people taking these measurements are somehow able to measure 0.1**μ**Hz on a 10**G**Hz scale?? Yeah, they're saving values like `5.9257052820783001 GHz`. Someone needs to teach them about significant figures... but that's beside the point. The point is I have to deal with this data :unamused:)

### Steps to Reproduce

1. Create the following json file:
    ```json
    {"foo": 4.4257052820783003}
    ```
2. Run pretty-format-json on it.

### Expected Output:

```json
{
  "foo": 4.4257052820783003
}
```

### Actual Output:

```json
{
  "foo": 4.4257052820783
}
```

The diff from expected is:

```diff
 {
-  "foo": 4.4257052820783003
+  "foo": 4.4257052820783
 }
```

### Version Info

+ pre-commit: 2.19.0
+ pre-commit-hooks: v4.2.0
+ Python: 3.8.8
+ OS: Debian 11

### Discussion

This might be something that has to be fixed within the python builtin `json` package. A custom JSON encoder/decoder that wraps things using `decimal.Decimal` might work too.

I've created a test case for this. See my [high-precision-numbers](https://github.com/dougthor42/pre-commit-hooks/tree/high-precision-numbers) branch or [the diff](https://github.com/pre-commit/pre-commit-hooks/compare/main...dougthor42:high-precision-numbers).

I'll see if I have time to actually fix this, but I don't expect to :frowning_face:.

asottile:
yeah I suspect this isn't really fixable due to `json` -- but feel free to take a stab at it

liortct:
As @asottile said the problem is in ```json``` standard library.
![image](https://user-images.githubusercontent.com/38353773/185232684-f241ea61-2ef1-4f93-a7c1-49bd873d09d5.png)
Therefore, this issue isn't fixable in the pre-commit hook and I advice to close this issue.

dougthor42:
@asottile How would you feel about a custom json encoder/decoder that handles floats using `decimal.Decimal`?

`json.loads` also has a `parse_float` arg that might be usable instead or in addition.

riccardobucco:
[...]

PR #818, opened 2022-10-06
Fix too long floating point number parse problem

SilverTux:
- Also include @dougthor42's diff to test the fix
    see: [his diff](https://github.com/pre-commit/pre-commit-hooks/compare/main...dougthor42:high-precision-numbers)
  - Soution for #780


pre_commit_hooks/pretty_format_json.py
```diff
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import argparse
-import json
 import sys
 from difflib import unified_diff
 from typing import Mapping
 from typing import Sequence
 
+import simplejson as json
```
asottile, line 9: sorry, don't want to pull in any more dependencies

PR #819, opened 2022-10-08
update readme for requirement files using custom names

midse:
Small improvement to describe how to sort requirement files with custom names.


asottile:
this is incorrect, the `files` filter is the standard way to control what files are matched by hooks

PR #824, opened 2022-10-18
Add `check-yaml-filename-extension` hook

taoufik07:
When working with YAML files, people tend to mix `.yaml` and `.yml`, and it goes unnoticeable : /


asottile:
also hooks should not modify the staging area (which `git mv` does) as this can break partially staged things -- you probably want to rename and use `--intent-to-add`

taoufik07:
> in the future please discuss features before spending time on them

Apologies, it just bugged me last night and I rapidly went to make this hook.

> this hook isn't really universally useful as many tools pick one convention and do not support both (perhaps ironically pre-commit would break if you applied your tool)

When not excluded and the wrong extension is passed, then ironically yes 😅.

Tools with one convention support can be excluded. The problem raises when a tool support multiple conventions or when the file is manually passed as input e.g. ansible, github actions, prettier, local files...

> also this can be done entirely using no-code via language: fail

I forgot about it, I'll do it locally then :D

PR #826, opened 2022-10-19
Forbid the combination of unique and ignore case options in file_contents_sorter.py

### LINKED ISSUE #794: `file-contents-sorter` Not working same way everytime
I get strange issue that file-contents-sorter hook does not work same way everytime. I made little test case which trickers the issue. If i run this 10 times it will sometimes pass and sometimes not. I did not have time to look what problem might be.

Add this to `file_contents_sorter_test.py`
```
(
    b'pre\nPre\n',
    ['--unique', '--ignore-case'],
    PASS,
    b'pre\nPre\n',
),
```

And run it with
`for run in {1..10}; do pytest tests/file_contents_sorter_test.py; done`
This gives me that about 30% pass and 70% fails

renegaderyu:
Example output when both options are specified:

```
usage: file_contents_sorter.py [-h] [--ignore-case | --unique] filenames [filenames ...]
file_contents_sorter.py: error: argument --unique: not allowed with argument --ignore-case
```

Closes #794.

### DIFF
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -49,25 +49,25 @@ def sort_file_contents(
 
 
 def main(argv: Sequence[str] | None = None) -> int:
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(exit_on_error=False)
     parser.add_argument('filenames', nargs='+', help='Files to sort')
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
         '--ignore-case',
         action='store_const',
         const=bytes.lower,
         default=None,
         help='fold lower case to upper case characters',
     )
-    parser.add_argument(
+    group.add_argument(
         '--unique',
         action='store_true',
         help='ensure each line is unique',
     )
-    args = parser.parse_args(argv)
-
-    if args.ignore_case and args.unique:
-        print('ERROR: usage of --unique and --ignore-case is unsupported. \
-        Please update your configuration.')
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/file_contents_sorter.py
```diff
@@ -65,6 +65,9 @@ def main(argv: Sequence[str] | None = None) -> int:
     )
     args = parser.parse_args(argv)
 
+    if args.ignore_case and args.unique:
+        return FAIL
```
asottile, line 69: this "passes the test" but now silently fails with no error message

renegaderyu: @asottile after reviewing and getting testing working locally I settled on removing the last test in `tests/file_contents_sorter_test.py` since the new expectation would be that file contents are unmodified, the errors is printed, and the check fails. Please let me know if this is sufficient or if you would like any other modifictions. Thanks.

pre_commit_hooks/file_contents_sorter.py
```diff
@@ -65,6 +65,11 @@ def main(argv: Sequence[str] | None = None) -> int:
     )
     args = parser.parse_args(argv)
 
+    if args.ignore_case and args.unique:
+        print('ERROR: usage of --unique and --ignore-case is unsupported. \
+        Please update your configuration.')
+        return FAIL
```
asottile, line 71: don't use backslashes like this -- also this is going to render very strangely

also this behaviour needs to be tested (`capsys` might help) -- but then again you can probably just make this use argparse's mutually exclusive options

PR #828, opened 2022-10-21
Work around ruamel parser error

### LINKED ISSUE #827: Work around upstream bug in ruamel-yaml
When adding a local tarball with PNPM, the YAML file `pnpm-lock.yaml` cannot be parsed with pre-commit-hook's YAML parser (ruamel) as it is currently instantiated by pre-commit.

```
  file:vendor/my-package.tgz:
    resolution: {integrity: sha512-pDhbxDJeZreZORpBacsEwMkJZtmnqwxDU5knDfLIf1l3YIzUggiIe1P2bz5QHu2mK5giSE3qgoZl28JHBoogTQ==, tarball: file:vendor/my-package.tgz}
    name: my-package
    version: 1.3.37
    dev: false
```

See https://sourceforge.net/p/ruamel-yaml/tickets/391/

The fix is simple; I will open a PR shortly.

hut8:
This bug made it impossible to check all pnpm-lock.yaml files, if those contained local package tarballs.

I checked this on my production pnpm-lock.yaml file which was how I found the bug; it works fine now, along with some other random YAML files.

Resolves #827

### DIFF
--- a/pre_commit_hooks/check_yaml.py
+++ b/pre_commit_hooks/check_yaml.py
@@ -8,7 +8,7 @@ from typing import Sequence
 
 import ruamel.yaml
 
-yaml = ruamel.yaml.YAML(typ='safe')
+yaml = ruamel.yaml.YAML(typ='safe', pure=True)
 
 
 def _exhaust(gen: Generator[str, None, None]) -> None:


asottile:
this is the slow loaded

Issue #273, opened 2018-03-19
Check-yaml exception for Ansible Vault


blackillzone:
Hi,

I've recently used the [embed encrypt variables](https://docs.ansible.com/ansible/latest/vault.html#encrypt-string-for-use-in-yaml) in my Ansible repos, and the check-yaml hook is failing because of the specific `!vault` usage.

```
Check Yaml...............................................................Failed
hookid: check-yaml

could not determine a constructor for the tag '!vault'
  in "inventories/group_vars/all.yml", line 62, column 13
```

Do you think it's possible to add this as an exception inside the hook (and make it pass for everyone who's using encypted variables in their playbooks) ?

Or should I add explicit exception inside my `.pre-commit-config.yaml`, for each files where I use those encrypted variables ?

asottile:
Might be with adding a "check-yaml-unsafe" which just attempts to validate syntax (and thus allowing extensions and unsafe constructs) instead of loading. Wouldn't be able to do some of the more advanced ideas I have for check-yaml such as duplicate key checking but it would enable this use case.

Thoughts?

asottile:
Made a patch for this, please try it out!  #274 (via `args: [--unsafe]`)

blackillzone:
Working like a charm for me and this use case ! Thank you for this quick patch :).

Looking at the PR, there is a small type error on line 55 in the `README.md`, and on line 45 in `pre_commit_hooks/check_yaml.py` : 
```
--unsafe` - Instaed of loading the files, simply parse them for syntax.
```

asottile:
This has been (finally!) released as part of [v1.3.0](https://github.com/pre-commit/pre-commit-hooks/releases/tag/v1.3.0)! 🎉

ssbarnea:
I am not sure that is the best approach but thanks for implementig that workaround.

asottile:
@ssbarnea do you have any suggestions?

ViViDboarder:
This is old, but the ability to ignore a single line would be one way to handle this and is a fairly common feature for linters.

asottile:
@ViViDboarder did you try the solution that's been released above in 2018?

ViViDboarder:
Yes, that works but it apparently prevents the other handy features such as checking for duplicate keys, no?

asottile:
yeah, yaml loading is all-or-nothing

PR #845, opened 2022-12-02
yaml: add new option to ignore "!ansible" tags

adrianlzt:
This new option for the yaml checker modify ruamel to remove the "!" from the "!vault" tag if it is found.
Removing that part allows the file to be parsed correctly, so other errors could be found.

fixes: #273

### DIFF
--- a/pre_commit_hooks/check_yaml.py
+++ b/pre_commit_hooks/check_yaml.py
@@ -65,15 +65,19 @@ def main(argv: Sequence[str] | None = None) -> int:
     load_fn = LOAD_FNS[Key(multi=args.multi, unsafe=args.unsafe)]
 
     if args.ignore_ansible_vault:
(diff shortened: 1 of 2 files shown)


asottile:
there is already an option: `--unsafe`

adrianlzt:
I have a global configuration for pre-commit, so adding ``--unsafe`` will apply to all other yaml files, so I prefer to avoid it.

asottile:
good news, you can specify `check-yaml` as many times as you want and with whatever options you want

adrianlzt:
Well, in case anyone is interested:
```
-   repo: https://github.com/adrianlzt/pre-commit-hooks
    rev: 1.0.0
    hooks:
    -   id: check-yaml
        args: ["--ignore-ansible-vault"]
```

asottile:
sorry this is not a platform to advertise

PR #847, opened 2022-12-05
feat: add whitelist support for detect-private-key hook

carlosbustillordguez:
The motivation for this PR is to add an opinion to whitelist files during the execution of the `detect-private-key` hook. For the following content the `detect-private-key` hook return the exit code `1`, which is a false positive:

```yaml
extraTls: []
  ## @param ingress.secrets Custom TLS certificates as secrets
  ## NOTE: 'key' and 'certificate' are expected in PEM format
  ## NOTE: 'name' should line up with a 'secretName' set further up
  ## If it is not set and you're using cert-manager, this is unneeded, as it will create a secret for you with valid certificates
  ## If it is not set and you're NOT using cert-manager either, self-signed certificates will be created valid for 365 days
  ## It is also possible to create and manage the certificates outside of this helm chart
  ## Please see README.md for more information
  ## e.g:
  ## secrets:
  ##   - name: rabbitmq.local-tls
  ##     key: |-
  ##       -----BEGIN RSA PRIVATE KEY-----
  ##       ...
  ##       -----END RSA PRIVATE KEY-----
  ##     certificate: |-
  ##       -----BEGIN CERTIFICATE-----
  ##       ...
  ##       -----END CERTIFICATE-----
```

This PR defines a new argument for the detect-private-key` hook, e.g.:

```yaml
- id: detect-private-key
    args: [--whitelist=.detect-private-key.whitelist]
```

Then the `.detect-private-key.whitelist` file can have the following content:
```
modules/kubernetes-rabbitmq-cluster/templates/values-rabbitmq-production.yaml.tpl
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,14 @@ The following arguments are available:
 #### `detect-private-key`
 Checks for the existence of private keys.
 
+The following arguments are available:
+- `--whitelist` - the filename with with the files (relative path) to whitelist. For example:
+
+  ```yaml
+  - id: detect-private-key
+    args: [--whitelist=.detect-private-key.whitelist]
+  ```
+
 #### `double-quote-string-fixer`
 This hook replaces double quoted strings with single quoted strings.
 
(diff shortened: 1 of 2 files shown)


carlosbustillordguez:
> `whitelist` is not appropriate language
> 
> `pre-commit` makes this unnecessary -- use the intended `exclude` mechanism

Hi @asottile! Thanks for pointing it out... I forgot completely the `exclude` mechanism with `pre-commit`.

PR #850, opened 2022-12-14
Stop pkg_resources from being added as well

### LINKED ISSUE #1030: requirements-txt-fixer doesn't handle new pkg_resources==0.0.0
Perhaps something changed on PyPI but instead of Ubuntu 20.04 generating `pkg-resources==0.0.0` it now generates `npkg_resources==0.0.0` which should also be ignored.

ericfrederich:
This package with an underscore instead of a dash seems to be new behavior (at least on Ubuntu 20.04 with Python 3.8

Fixes #1030

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -115,7 +115,10 @@ def fix_requirements(f: IO[bytes]) -> int:
     # which is automatically added by broken pip package under Debian
     requirements = [
         req for req in requirements
-        if req.value != b'pkg-resources==0.0.0\n'
+        if req.value not in [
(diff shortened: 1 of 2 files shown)


ericfrederich:
I'd appreciate this merge.  I'm not sure if something changed on PyPI or Debian or Ubuntu, but now on my Ubuntu machine I'm getting `pkg_resources` with an underscore instead of `pkg-resources` like I used to get.

asottile:
can't merge without a test demonstrating the new behaviour

ericfrederich:
My apologies, 1.5 years later I added tests and a corresponding issue to close as well.

asottile (approved):

PR #856, opened 2023-01-12
Add `--exclude-values` arg to `detect-aws-credentials`

sapslaj:
Adds an argument to pass a regex which will exclude any values that match it to not count as a failure.

The rationale behind this is test variables, such as those used with [LocalStack](https://localstack.cloud/) (in some cases literally `TEST`!), are counted as a failure. Adding this parameter will allow bypassing the check for those values only instead of excluding the files from the hook running where _real_ credentials could potentially leak.

Happy to discuss alternative approaches to this problem; this was just my first stab at tackling it.

### DIFF
--- a/pre_commit_hooks/detect_aws_credentials.py
+++ b/pre_commit_hooks/detect_aws_credentials.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import configparser
 import os
+import re
 from typing import NamedTuple
 from typing import Sequence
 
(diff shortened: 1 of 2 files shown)


asottile:
please discuss features before working on them

`exclude` is intended for this

sapslaj:
@asottile What's the appropriate workflow for proposing features? I don't see a `CONTRIBUTING.md` or anything of the sort.

Also, I explained how `exclude` was undesirable unless there's a way to use `exclude` to exclude contents of files and not filenames.

asottile:
it's not different from open source in general -- open an issue describing your problem before jumping to solutions

PR #863, opened 2023-01-19
Make it clear that this isn't a real credential

katzj:


asottile:
this doesn't improve anything -- it's a test

asottile:
also "dummy" is ableist language -- tend to prefer "placeholder"

katzj:
It is a test but it avoids a scanner accidentally seeing it and deciding it's a real credential and flagging it. Happy to change it say placeholder instead though (good callout, thanks)

asottile:
"a scanner"?

PR #867, opened 2023-01-24
Feature/check for iam

brycewade1:
Add check-iam hook


asottile:
please discuss features first before spending time on them

brycewade1:
So embarrassing.  That was supposed to be PR for my local main for
testing.  Good thing you have good tests to catch my bad code!

PR #873, opened 2023-02-16
feat(check-added-large-files): :sparkles: add --no-skip-lfs-files arg…

JulienGrv:
…ument

### DIFF
--- a/pre_commit_hooks/check_added_large_files.py
+++ b/pre_commit_hooks/check_added_large_files.py
@@ -34,13 +34,16 @@ def find_large_added_files(
         filenames: Sequence[str],
         maxkb: int,
         *,
+        skip_lfs_files: bool = False,
         enforce_all: bool = False,
 ) -> int:
     # Find all added files that are also in the list of files pre-commit tells
     # us about
     retv = 0
     filenames_filtered = set(filenames)
-    filter_lfs_files(filenames_filtered)
+
+    if skip_lfs_files:
+        filter_lfs_files(filenames_filtered)
 
     if not enforce_all:
         filenames_filtered &= added_files()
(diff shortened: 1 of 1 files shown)


JulienGrv:
Hi @asottile is there something missing in my pull request or is it just you don't want to add this feature?

asottile:
it looked like spam or that you didn't mean to open a PR

JulienGrv:
I forgot to specify a description. Would it be possible to have such a feature?

asottile:
usually _features_ start with a discussion about the use case -- and after searching the issue tracker to find duplicates

and we don't do emoji or conventional commit garbage

JulienGrv:
I will remove the conventional commit, I didn't mean to offend you with it.

Usually for such small change I just open a PR and discuss it in the PR but I can also start a discussion if you estimate it is necessary. The feature is to have a flag to skip the filtering of git lfs files in the check-added-large-files hook.

Should I open another PR without conventional commit or should I open a discussion first?

asottile:
I understand _what_ but a feature needs a _why_ and tests demonstrating that _why_

JulienGrv:
Ok, now that you point this out with the _why_, I realize that this hook is not a suitable place for this feature because it is a hook for non-lfs files and including lfs files wouldn't make sense. For LFS files, the size limit could be actually bigger. In that case what would make more sense for my use case is to have a different hook which only focuses on LFS files. I will explore this possibility and write my own hook

asottile:
indeed -- and this is why it is always best to start with an issue for features, such that you have a discussion realizing that the problem statement you have makes no sense and that jumping to a solution is a waste of everyone's time

PR #892, opened 2023-04-19
end_of_file_fixer: write platform native line endings

bwncp:
The most common setup for git is to use platform-native line endings in checkout (`core.autocrlf` set to `true`). So the eol file fixer should write native line endings.


bwncp:
I can add a command line option so that the behavior can be configured. Or read the git option in the hook?

asottile:
this isn't right -- it should use the line ending of the file

PR #896, opened 2023-04-20
Detect import rdb from celery.contrib

q0w:


q0w:
but 'rdb' does not detect at all rdb usage

PR #897, opened 2023-04-20
end_of_file_fixer: detect line ending

bwncp:
Instead of writing \n we try to detect the line ending used in the file. We use the first line ending sequence we find.

### DIFF
--- a/pre_commit_hooks/end_of_file_fixer.py
+++ b/pre_commit_hooks/end_of_file_fixer.py
@@ -6,6 +6,26 @@ from typing import IO
 from typing import Sequence
 
 
+def detect_eol_equence(file_obj: IO[bytes]) -> bytes:
+    # readline() doesn't work because it doesn't get \r right
+    eol_marker = b''
+    last_was_eol = False
+    while True:
+        next = file_obj.read(1)
+        if not next:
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/end_of_file_fixer.py
```diff
@@ -6,6 +6,26 @@
 from typing import Sequence
 
 
+def detect_eol_equence(file_obj: IO[bytes]) -> bytes:
+    # readline() doesn't work because it doesn't get \r right
+    eol_marker = b''
+    last_was_eol = False
+    while True:
+        next = file_obj.read(1)
```
asottile, line 14: this is going to be extremely slow

asottile: I don't have one, just i know this isn't acceptable

asottile: you've got IO[bytes] so no buffering

bwncp: I think that is just the type annotation and doesn't change the buffering of objects returned by `open`. And I run `strace`.

But maybe you have some documentation?

PR #919, opened 2023-06-14
Add optional --quiet arg to trailing_whitespace

NickLupien-Silk:
Allows `trailing_whitespace` to simply run without creating a failure. example:
```
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.4.0
    hooks:
      - id: trailing-whitespace
        args: [--quiet]
```


asottile:
this does not do what you think it does

PR #924, opened 2023-06-27
Add option to reject requirements files if any packages are missing a version string

rbm-radius:
Rationale: some teams require specifying a version for every package listed in their requirements files, for quasi-reproducible builds that will prevent new versions of pip packages breaking their code.

This option is off by default, to preserve backward compatibility.


asottile:
please discuss features before spending time on them

this is out of scope for this tool.  it's trivial to make a pygrep hook as well if you don't want to write any code

PR #928, opened 2023-07-05
add __import__("pdb") handling

aless10:
Sometimes, we use the one liner `__import__('pdb').set_trace()` to set the breakpoints which was not supported.
This PR aims to handle this case, so that we can use the `debug_statement_hook` and catch this.


asottile:
this is intentionally left as an escape hatch

PR #929, opened 2023-07-05
add __import__("pdb") handling

aless10:
This PR aims to add the support for the `__import__("pdb").set_trace()` one liner. 
Since we use the debug-statement-hook and also this type of pdb import, I thought it can be useful.


aless10:
Sorry, I did not see you comment

PR #930, opened 2023-07-07
Move file-contents-sorter arguments

dupuy:
These were in the wrong place (maybe a botched merge?)


asottile:
ah shoot I didn't see this one and should have merged it instead of #931 . my bad

PR #945, opened 2023-08-13
for pretty-format-json add option: --empty-object-with-newline

ascheucher-shopify-partner:
In Shopify theme development, there is a GitHub bot which pulls in changes in GitHub repos.
Sadly it formats .json files in a non standard way. Hence, every empty object get's a newline, which doesn't not conform to standard formatting. 
This creates then a new commit in the branch, which is pushed back into the origin repo. Which needs to be pulled to be able to push again. It's creating a cycle.

Hence, I created the formatting option _--empty-object-with-newline_ to conform to the Shopify .json formatting to prevent commit pollution.

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -2,16 +2,17 @@ from __future__ import annotations
 
 import argparse
 import json
-import sys
 import re
-
+import sys
 from difflib import unified_diff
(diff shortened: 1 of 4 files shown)


asottile:
please discuss features before spending time on them.  this is far too specific to be added and supported

asottile:
you should fix the other thing rather than adding hacks here

PR #961, opened 2023-09-06
Document newline behavior of `file-contents-sorter`

Maetveis:
Let's document the behavior and suggest what to do for users of the hook.

Ref: #958, #959

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Makes sure files end in a newline and only a newline.
 Sort the lines in specified files (defaults to alphabetical).
 You must provide the target [`files`](https://pre-commit.com/#config-files) as input.
 Note that this hook WILL remove blank lines and does NOT respect any comments.
+All newlines will be converted to line feeds (`\n`).
 
 The following arguments are available:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,11 +117,20 @@ Makes sure files end in a newline and only a newline.
...
 - `--unique` - ensure each line is unique.
 
+When using this hook in a git repository on Windows it is recommended to set
+[`core.autocrlf`](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_autocrlf)
+to `false` or set the end-of-line to line feed via
+[`gitattributes`](https://www.git-scm.com/docs/gitattributes#_eol) for the
+targeted files. Without these changes the hook will fail, but produce no
+apparent change in `git diff`. See
+[#958](https://github.com/pre-commit/pre-commit-hooks/issues/958) for details.
+
```
asottile, line 133: I would just leave the first part about `\n` -- we don't need a whole tutorial about newlines here

Maetveis: Okay, I reverted this part.

PR #962, opened 2023-09-06
Add detect_secret_token

### LINKED ISSUE #940: Feature request: Check for RFC 8959 "secret-token" scheme
[RFC 8959](https://datatracker.ietf.org/doc/html/rfc8959) registers the `secret-token:` URI scheme with the intent of establishing a common syntax for authentication tokens and making it easier to detect tokens and prevent them from leaking into e.g. version control systems.

Seems a natural fit for this project, given it already has checks for private keys and AWS credentials.

mxr:
Resolves #940

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -119,6 +119,12 @@
     entry: detect-private-key
     language: python
     types: [text]
+-   id: detect-secret-token
+    name: detect secret token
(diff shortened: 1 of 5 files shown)


asottile:
#940 should probably be rejected, this can be implemented trivially with `pygrep`

mxr:
Good call

PR #965, opened 2023-09-21
Remove redundant int() conversion

eumiro:


asottile (approved):
this is leftover from python 2 where `math.ceil` returned a float

Issue #960, opened 2023-09-06
`requirements-txt-fixer` does not remove duplicated elements


csibe17:
In `requirements-txt-fixer` if there are more than 1 requirement with the same params, the duplicates are not removed.

asottile:
can you show an example?

csibe17:
```
dependency1==1.0.0
dependency2==1.0.0
dependency3==1.0.0
dependency1==1.0.0
```

will end up as

```
dependency1==1.0.0
dependency1==1.0.0
dependency2==1.0.0
dependency3==1.0.0
```

JaySharmaSdsu:
I am totally new to contributing on open source. If this is indeed an issue. Can I work on it?'

asottile:
sure go for it, let me know if you need any help

JaySharmaSdsu:
Sick! Excited!! Let me get back if I need any help.

PR #968, opened 2023-09-27
Remove duplicated elements

HeySlava:
I do not remove duplicated comments, only values

#960


asottile:
someone else is already working on #960

this also makes the hook do N**2 work which is worse

PR #979, opened 2023-10-19
[no_commit_to_branch] Add user friendly error message

Destroy666x:
Prints a user-friendly message in case the action is disallowed on a branch. As of now it just shows "error", which is fine if you added the hook, but terrible UX for people using it in some project (e.g. installed automatically in devcontainer) for the 1st time.

### DIFF
--- a/pre_commit_hooks/no_commit_to_branch.py
+++ b/pre_commit_hooks/no_commit_to_branch.py
@@ -9,29 +9,37 @@ from pre_commit_hooks.util import CalledProcessError
 from pre_commit_hooks.util import cmd_output
 
 
-def is_on_branch(
-        protected: AbstractSet[str],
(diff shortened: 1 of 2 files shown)


asottile:
please discuss issues before spending time on a pull request -- thanks

asottile:
you can override `name` to whatever makes sense for your usecase

Destroy666x:
Overriding `name` in a completely non-dynamic way as a universal recipe for fixing UX is really not it and I expected that kind of comment to appear looking at what I found in your repos - e.g. one issue which bunch of downvotes for your comments stating that custom error messages for hooks would be useless.

Also, I spent time on it mostly for people who'd like an alternative - they can feel free to use the fork like I'll do, ¯\_(ツ)_/¯ what happens with this PR - you made a quick choice which is fine with me.

PR #987, opened 2023-10-31
Fix sort_simple_yaml for comments with leading spaces

r-downing:
The `sort_simple_yaml.py` script doesn't handle comments correctly if they have any space before the '#'. Added a test-case to demonstrate, and a simple fix


asottile:
this feels like a garbage in garbage out scenario

the tool only supports top level yaml so supporting indentation is intentionally omitted

PR #993, opened 2023-11-20
add check-yaml-sorted script and tests

r-downing:
The existing `sort-simple-yaml` check seems to only handle **really** simple yaml files, and can potentially destroy those with nested objects or comments, etc...

I wanted an alternative that would work with more complex yaml files, without corrupting them, that just performs a check that the top-level is sorted.

This does so by iterating through the top-level object, whether it's a list, dict, etc, and comparing the sorting based on string representations. This lets it handle all kinds of configurations, such as

Examples: 

```yaml
- a
- simple
- list
```

would be

```yaml
- a
- list
- simple
```

---

```yaml
dicts: ...
by: ...
keys: ...
```

becomes

```yaml
by: ...
dicts: ...
keys: ...
```

---

```yaml
- list: ...
- of: ...
- dicts: ...
```

would be 

```yaml
- dicts: ...
- list: ...
- of: ...
```

---

```yaml
- name: list
  field2: of
- name: dicts
  field2: with
- name: same
  field2: keys
- name: by
  field2: first value
```

would be

```yaml
- name: by
  field2: first value
- name: dicts
  field2: with
- name: list
  field2: of
- name: same
  field2: keys
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -95,6 +95,12 @@
     entry: check-yaml
     language: python
     types: [yaml]
+-   id: check-yaml-sorted
+    name: check yaml sorted
+    description: checks top-level items of yaml files are sorted.
+    entry: check-yaml-sorted
+    language: python
+    types: [yaml]
 -   id: debug-statements
     name: debug statements (python)
     description: checks for debugger imports and py37+ `breakpoint()` calls in python source.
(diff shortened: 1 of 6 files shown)


asottile:
yeah it's intentionally only extremely simple yaml as documented

r-downing:
> yeah it's intentionally only extremely simple yaml as documented

Yeah that's why I made this alternative option that could handle all kinds of files, thought it might be useful?

Issue #997, opened 2023-11-30
[bug] end-of-file-fixer doesn't handle last line with spaces


r-downing:
Not sure if this is intentional, but the end-of-file fixer adds an extra line if the final line has spaces. Technically, combined with the `trailing-whitespace` check and running two passes, it would eventually fix it, but since it's called `end-of-file-fixer` perhaps it should handle this case on its own? I could put up a PR if you think it's worth fixing

File with spaces on last line:
![image](https://github.com/pre-commit/pre-commit-hooks/assets/26515643/59e204e5-a73f-4bb5-9415-90457c1a170e)

After running `end-of-file-fixer` (added a line):
![image](https://github.com/pre-commit/pre-commit-hooks/assets/26515643/a6037880-35fa-452e-bfb5-cbc40d303025)

With pre-commit hooks in this order:
```yaml
    hooks:
      - id: end-of-file-fixer
      - id: trailing-whitespace
```
You would actually have to run and stage _twice_. It would first add the **extra** line, then the trailing space would be removed from the second-to-last line, so you'd end up with 2 newlines at the end

After running with first pass:
![image](https://github.com/pre-commit/pre-commit-hooks/assets/26515643/5abc06b2-7278-4b53-aa2a-132da623c1e9)

asottile:
end-of-file-fixer intentionally only handles newlines

garbage in, garbage out -- you should also use `trailing-whitespace`

r-downing:
Perhaps a warning in the readme that `trailing-whitespace` has to be run **first**?

PR #998, opened 2023-11-30
add note for `end-of-file-fixer` to run `trailing-whitespace` first, so you don't have to run and stage twice

r-downing:
See https://github.com/pre-commit/pre-commit-hooks/issues/997 for example


asottile:
this doesn't really matter -- it will resolve in two passes just fine

PR #1015, opened 2024-02-26
feat: add conventional commit check

katomaso:
Add simple pygrep expression for basic checking of conventional commit message in `commit-msg` stage


asottile:
please discuss features before wasting time on them

there are many better tools for this that alreadyv exist

plus I personally think conventional commits are a huge waste of time

PR #1025, opened 2024-03-14
pretty-format-json: ignore original newline presence or absence

ajoga:
I propose to have the `pretty-format-json` hook not care/change the end of file line separator's presence.

If  there's a newline, we keep it, if there's no newline, we don't add it.

This is motivated by:
1. the [default setting for json formatter in vscode](https://code.visualstudio.com/docs/getstarted/settings#_default-settings) that drops any endline caracter (`"json.format.keepLines": false`)
2. the inintuitive behavior for our existing json files that we created with a sole `json.dumps()` that tripped this hook
3. the redundancy with existing `end-of-file-fixer` hook, for which I ironically had to add an exception in this commit :)

I could make it as an argument if you want, but it feels overkill as I'm writing this. let me know


asottile:
not ending with a newline is wrong so this will not be accepted

please search for duplicates next time as this has already been rejected at least once

PR #1031, opened 2024-03-26
Add check for illegal Widnows names

### LINKED ISSUE #589: Add check for illegal Windows filenames

ericfrederich:
Fixes #589

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -40,6 +40,11 @@
     language: python
     types: [text, executable]
     stages: [commit, push, manual]
+-   id: check-illegal-windows-names
+    name: check illegal windows names
+    entry: Illegal windows filenames detected
+    language: fail
+    files: '(?i)(^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(\.|/|$)'
 -   id: check-json
     name: check json
     description: checks json files for parseable syntax.
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Checks for a common error of placing code before the docstring.
 #### `check-executables-have-shebangs`
 Checks that non-binary executables have a proper shebang.
 
+#### `check-illegal-windows-names`
+Check for files that cannot be created on Windows.
+
 #### `check-json`
 Attempts to load all json files to verify syntax.
 
--- a/tests/check_illegal_windows_names_test.py
+++ b/tests/check_illegal_windows_names_test.py
@@ -1,0 +1,46 @@
+from __future__ import annotations
+
+import os.path
+import re
+
+import pytest
+
+from pre_commit_hooks.check_yaml import yaml
+
+
+@pytest.fixture(scope='module')
+def hook_re():
+    here = os.path.dirname(__file__)
+    with open(os.path.join(here, '..', '.pre-commit-hooks.yaml')) as f:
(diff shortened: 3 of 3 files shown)


ericfrederich:
Took advice from #590 and created a new PR ... sorry for delay

.pre-commit-hooks.yaml
```diff
@@ -40,6 +40,11 @@
     language: python
     types: [text, executable]
     stages: [commit, push, manual]
+-   id: check-illegal-windows-names
+    name: check illegal windows names
+    entry: Illegal windows filenames detected
+    language: fail
+    files: '(?i)(^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(\.|/|$)'
```
asottile, line 47: can we add a quick little test which loads this regex from the configuration and runs it against a few expected values?

ericfrederich:
Tests added, let me know if that's what you had in mind.

asottile (approved):

.pre-commit-hooks.yaml
jleclanche: @asottile @ericfrederich I recommend checking for the `:` character as well (see my comment in #589 regarding :Zone.Identifier files).

ericfrederich: @jleclanche looks like this got merged without code checking for the `:` but no release has been made yet.

I just checked and it's definitely a problem on Windows.  I created a file `foo:bar` on linux, commited, pushed, then pulled down the repo from Windows and got greeted with `error: invalid path 'foo:bar'`

I'll add a test for it and a new PR

thetillhoff:
Any plans on releasing a new version soon with this check included?

Just read about it on the Webpage, tried to add it, but it wasn't found. Then checked the dates, and the last release of pre-commit was before this PR was merged...

asottile:
you're confused?

thetillhoff:
Am I?
This PR adds the check-illegal-windows-names hook, and was merged on April 16th.
The last released version of this repo (https://github.com/pre-commit/pre-commit-hooks/releases/tag/v4.6.0) was released on April 6th.

Is there a plan to create a new release that includes the check-illegal-windows-names hook?
Or am I misunderstanding something?

asottile:
have you read the pre-commit docs?

thetillhoff:
Probably not the part you have in mind. You are welcome to point me to what you want me to read.

Issue #1032, opened 2024-04-02
Deprecate/remove `fix-encoding-pragma` hook?


mxr:
Since py2 is long unsupported

asottile:
I think yeah, or at least make the default behavior to remove it -- or perhaps recommend pyupgrade instead

mxr:
Suggesting pyupgrade makes sense to me. So to be clear the hook would no-op and fire a DeprecationWarning which suggests pyupgrade?

asottile:
probably a minor release which prints a warning to stderr and adding `(deprecated)` to the hook `name`

and then in 6 months a major which deletes it (autoupdate handles this nicely)

PR #1033, opened 2024-04-02
Add deprecation messaging for `fix-encoding-pragma`

mxr:
Related to #1032


asottile:
we should make sure to mention this in the README and probably explicitly mention the replacement

mxr:
Updated

asottile (approved):

PR #1039, opened 2024-04-12
Continues processing JSONs even if hook fails (fixes #1038)

### LINKED ISSUE #1038: `pretty-format-json` stops formatting JSONs when it encounters one with incorrect format (e.g. JSON5 comments)
Transferred from https://github.com/pre-commit/pre-commit/issues/3124.

`pretty-format-json` prints "Input File {json_file} is not a valid JSON, consider using check-json" when it encounters an incorrectly formatted JSON file (e.g. with JSON5 comments), but it's not clear from that message that it then stops further processing files.

As `pre-commit` can partition the processed files into batches, this makes it even harder to tell as multiple batches of files can fail midway through with a few "X is not a valid JSON" messages printed, giving the impression that all the JSONs were formatted (or attempted formatted) and only those files printed are incorrect. The reality is that some JSONs may not have been formatted. I think this hook should try to format every JSON and only fail at the end if any have an invalid JSON format.

amarvin:
Fixes #1038

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -115,16 +115,20 @@ def main(argv: Sequence[str] | None = None) -> int:
                 f'Input File {json_file} is not a valid JSON, consider using '
                 f'check-json',
             )
-            return 1
-
-        if contents != pretty_contents:
-            if args.autofix:
-                _autofix(json_file, pretty_contents)
-            else:
-                diff_output = get_diff(contents, pretty_contents, json_file)
-                sys.stdout.buffer.write(diff_output.encode())
-
             status = 1
+        else:
+            if contents != pretty_contents:
+                if args.autofix:
+                    _autofix(json_file, pretty_contents)
+                else:
+                    diff_output = get_diff(
+                        contents,
+                        pretty_contents,
+                        json_file,
+                    )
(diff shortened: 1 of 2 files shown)


amarvin:
@asottile, any chance you could consider this change, please?

pre_commit_hooks/pretty_format_json.py
```diff
@@ -115,7 +115,8 @@ def main(argv: Sequence[str] | None = None) -> int:
                 f'Input File {json_file} is not a valid JSON, consider using '
                 f'check-json',
             )
-            return 1
+            status = 1
+            pretty_contents = contents
```
asottile, line 119: this should use `try` `else`

asottile (commented):
this needs a test which is why I didn't look at it

amarvin:
> this needs a test which is why I didn't look at it

Great point! I added a test that failed in `main`, but passes in this branch.

tests/pretty_format_json_test.py
```diff
@@ -82,6 +82,28 @@ def test_autofix_main(tmpdir):
     assert ret == 0
 
 
+def test_invalid_main(tmpdir):
```
amarvin, line 85: Could also consider combining this test with the one above.

asottile (approved):

PR #1042, opened 2024-04-15
feat: process fstring when python version gte 3.12

iFurySt:
image


asottile:
if you check the history of this file it is intentionally skipped

PR #1049, opened 2024-04-25
Extend check for illegal Windows filenames

### LINKED ISSUE #1048: Extend check for illegal Windows filenames
The `check-illegal-windows-names` hook does not check for every illegal filename on Windows. For example, filenames with "double" quotes are still allowed. 

Update the regex so that all illegal filenames get detected. Illegal Windows filenames are defined [here](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file?wt.mc_id=studentamb_357056).

Jeffrey-Lim:
Fixes #1048. It took some time to get it right, but I'm pretty sure this covers every possible illegal file name now. 

For reference, the following names are illegal on Windows:
- A name with any of the characters `< > : " \ | ? *` or any characters in the ASCII range `0-31`
- CON, PRN, AUX, NUL, COM0, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM¹, COM², COM³, LPT0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9, LPT¹, LPT², and LPT³
- A name ending with a space or period

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -44,9 +44,9 @@
     minimum_pre_commit_version: 3.2.0
 -   id: check-illegal-windows-names
     name: check illegal windows names
-    entry: Illegal windows filenames detected
+    entry: Illegal Windows filenames detected
     language: fail
-    files: '(?i)((^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(\.|/|$)|:)'
+    files: '(?i)((^|/)(CON|PRN|AUX|NUL|COM[\d¹²³]|LPT[\d¹²³])(\.|/|$)|[<>:\"\\|?*\x00-\x1F]|/[^/]*[\.\s]/|[^/]*[\.\s]$)'
 -   id: check-json
     name: check json
     description: checks json files for parseable syntax.
--- a/tests/check_illegal_windows_names_test.py
+++ b/tests/check_illegal_windows_names_test.py
@@ -30,6 +30,17 @@ def hook_re():
         pytest.param('com7.dat', id='com with digit'),
         pytest.param(':', id='bare colon'),
         pytest.param('file:Zone.Identifier', id='mid colon'),
+        pytest.param('path/COM¹.json', id='com with superscript'),
+        pytest.param('dir/LPT³.toml', id='lpt with superscript'),
+        pytest.param('with < less than', id='with less than'),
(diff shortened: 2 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -42,9 +42,9 @@
     stages: [commit, push, manual]
 -   id: check-illegal-windows-names
     name: check illegal windows names
-    entry: Illegal windows filenames detected
+    entry: Illegal Windows filenames detected
     language: fail
-    files: '(?i)((^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(\.|/|$)|:)'
```
asottile, line 47: at the very least the changes to `COM` and likely LPT do not seem correct:

```
C:\Users\Anthony\y>dir
 Volume in drive C has no label.
 Volume Serial Number is A6A6-1A8E

 Directory of C:\Users\Anthony\y

04/27/2024  02:27 PM    <DIR>          .
04/27/2024  02:27 PM    <DIR>          ..
04/27/2024  02:27 PM                 0 com0
04/27/2024  02:27 PM                 0 com¹
               2 File(s)              0 bytes
               2 Dir(s)  239,653,076,992 bytes free
```

Jeffrey-Lim: I've taken another at it. I am actually able to create files named `com0`, `lpt0`, and `clock$`. I was surprisingly also able to create illegal filenames if there was an extension, so `COM1.txt` works perfectly fine for me as well. However, I cannot make files named `com¹` or `lpt¹`. I'm curious how you were actually able to create `com¹`. 

I've seen in [this video](https://youtu.be/s1ct8VPB-4U?si=6PS_BHzFYBkNkKlq&t=751) that creating `COM1.txt` on Windows 10 was illegal and behaves weirdly if you do it anyway. But as I mentioned before, it worked fine for me on Windows 11 23H2. Perhaps different versions of Windows have different restrictions? I can imagine that the Windows devs are slowly getting rid of these restrictions, so maybe `COM1.txt` works in most use cases now, but still not in all. 

So just to be safe, I think it's fine to stick to [Microsoft's own documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions), even if it is slightly more restrictive than Windows actually is. I will remove `clock$` though since that is not actually listed there and it seems to only affect Windows NT and older.

asottile (approved):

PR #1072, opened 2024-07-10
feat: add a simple HTML check pre-commit hook

bernardpaulus:
Hello,

I needed a simple, basic validation for HTML files.

Here is a proposal for a pre-commit hook `check-html`.

Untested (I don't know how to run the unit tests locally) - let's see how it fares in the CI pipeline


asottile:
hi please create issues to discuss features before spending time on them

I don't think this makes sense here.  there are many pre-existing html validators (w3c comes to mind) which are going to do a much better job than python's html.parser

PR #1082, opened 2024-08-15
Add `--replace-single-quotes` to replace single quotes with double ones

### LINKED ISSUE #1083: Proposal: Allow `double-quote-string-fixer` to replace single quotes into double ones
Some projects prefer double quotes over single quotes. It is not hard to add a `--replace-single-quotes` to let string_fixer.py to do the reversed thing.

wjsi:
Fixes #1083.


asottile:
only supporting single quotes is intentional

PR #1089, opened 2024-09-02
fix pre-commit-2332

rishabhjain1712:
Added a fix for address the issue of inconsistent mixing between tab and space indentation within the JSON content.

### DIFF
--- a/pre_commit_hooks/check_json.py
+++ b/pre_commit_hooks/check_json.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 import argparse
 import json
 import re
-from typing import Any, Sequence
+from typing import Any
+from typing import Sequence
 
 
 def raise_duplicate_keys(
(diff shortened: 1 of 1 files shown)


asottile:
please in the future discuss changes before wasting time on them

rishabhjain1712:
got it, this was my first contribution so didnt know much about how to go about these things.
will keep this in mind

rishabhjain1712:
@asottile if you dont mind can you guide me a little on how to contribute

asottile:
generally it's best to start with an issue describing your problem before jumping to solutioneering.  make sure it's something that's agreed upon first before spending a bunch of effort trying to build it

also unrelated to that: use a commit subject to describe _what_ your change does rather than "fix pre-commit-2332" (which I don't even know what that's a reference to? issue 2332 doesn't exist yet?)

PR #1106, opened 2024-11-04
Allow display of custom message when no-commit-to-branch fails

ssbarnea:
This change should make the failures of no-commit-to-branch easier for people that might encounter it as maintainers can provide information about allowed branch names.

Current behavior just gives an opaque error that is not very useful for people that do not have a lot of knowledge about the project they are contributing to.


asottile:
the intention is you would use `name`

Issue #94, opened 2015-12-18
Hook to add license comments


asottile:

asottile:
I'm not sure it would need to be, depending on the scope here. for instance I would probably see this configured as:

```yaml
    -   id: add-license
        args: [--license, apache, --owner, 'Anthony Sottile', --copyright-year, '2018']
```

and then it would ensure comments at the top of files like [this](https://github.com/Yelp/paasta/blob/ea5c03be3382f762503381410890e76e03a4c916/paasta_tools/drain_lib.py#L1-L13)

jcpunk:
Personally, I'd love something that checks for `SPDX` tags.

asottile:
there are existing hooks for this in https://pre-commit.com/hooks.html

PR #1112, opened 2024-12-06
Add hook to check SPDX lines

marcusburghardt:
I used a local hook to check this line but it would be great to have it in `pre-commit` as well so this can also be useful in more projects.

I read the documentation, looked some other examples and tested it locally but this is my first PR here, so please let me know what should be changed to properly fit the project standards. I will leave the PR as Draft for now.

Thanks

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -61,6 +61,13 @@
     types: [text]
     stages: [pre-commit, pre-push, manual]
     minimum_pre_commit_version: 3.2.0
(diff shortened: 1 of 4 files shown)


marcusburghardt:
> hi thanks however this has been rejected before

I only found the https://github.com/pre-commit/pre-commit-hooks/issues/94 but didn't find an existing hook to check this. Could you provide me more details, please?

Also, is there any criteria or guidance about what can or cannot be included?

Thanks

PR #1113, opened 2024-12-11
Fix wrong target for sort-simple-yaml

ctcjab:
Configure sort-simple-yaml to apply to only yaml files.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -201,7 +201,7 @@
     description: sorts simple yaml files which consist only of top-level keys, preserving comments and blocks.
     language: python
     entry: sort-simple-yaml
-    files: '^$'
+    types: [yaml]
 -   id: trailing-whitespace
     name: trim trailing whitespace
     description: trims trailing whitespace.


asottile:
if you read the docs for this hook you'll understand why the default is the way it is

ctcjab:
Where are the docs you're referring to? I had already read the docstrings in https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/sort_simple_yaml.py and they all talk about YAML files, there is no mention that this hook is meant for other filetypes.

asottile:
the README is typically the place to look first

ctcjab:
> Note that sort-simple-yaml by default matches no files as it enforces a very specific format. You must opt in to this by setting files

I see what happened: Whoever set this hook up in the monorepo where I hit this did so via `files: ^our_team_subdirectory/` and did not include any extension matching in that regex (e.g. `[^.]+\.ya?ml$` at the end), so it ended up unintentionally rewriting .py files. Perhaps this person made the reasonable assumption that this hook was already limiting itself to yaml files.

I already fixed the regex in our repo (before even submitting this PR). I am still wondering if it makes sense to accept this PR – would including the `types: [yaml]` here not make it easier to use this hook?

PR #1131, opened 2025-01-30
upgrade asottile/workflows

asottile:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,12 +8,12 @@ on:
 
 jobs:
   main-windows:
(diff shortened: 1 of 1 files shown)


xavier2k6:
@asottile Apologies for the noise/off-topic, but how come Python 3.13/3.14 (`py3.13`/`py3.14`) aren't added to the `tox` matrix?

https://github.com/pre-commit/pre-commit-hooks/blob/ec458dc1db98861800bd35c16a7d97e8cdb61484/.github/workflows/main.yml#L18

xavier2k6:
Thank you for the feedback.

PR #1132, opened 2025-02-03
Check for web3 private keys

kongzii:
Hi, I'm adding a hook to check for the web3 private keys (keys used to sign transactions on the blockchain) in the repository.


asottile:
no thanks, we're generally not adding new hooks unless they're universally useful (and planet burning grift does not fall in that category)

PR #1138, opened 2025-02-21
Empty object with newline

ascheucher-shopify-partner:


asottile:
this does not seem like a good idea

PR #1142, opened 2025-03-05
Include check for a version specification into requirements_txt_fixer hook

binka-dev:
Hi all,

for my team it is best practice to specify the version of a requirement within the `requirements.txt` file. 
We do this mainly to prevent issues introduced by incompatible dependency updates.

This pull request aims to include a new feature within the `requirements-txt-fixer` hook to check for version specifications in requirements files. It fails if a dependency is listed without a version specified. This behavior is optional and can be turned on or off.

Feel free to adjust the code or provide feedback :)

Best regards

Yannik

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -192,7 +192,7 @@
     always_run: true
 -   id: requirements-txt-fixer
     name: fix requirements.txt
-    description: sorts entries in requirements.txt.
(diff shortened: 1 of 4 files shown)


binka-dev:
@asottile Thank you for your feedback. Don't be worried that I wasted time. 
It is useful for our team beside of the probably incorrect naming. 

Do you think such a functionality would be interesting for this repository as part of an independent "linter"?
I think it is useful in general and didn't find a comparable hook so far.

PR #1148, opened 2025-03-18
[pretty_format_json] Add compact array feature

adithyabsk:
- Find non-nested numeric arrays using JSON spec: https://www.json.org/json-en.html

Note: in the process of testing this patch, I noticed that we have a round trip related bug when using `json.loads` followed by `json.dump` when dealing with small numbers (which get converted to scientific notation) and some scientific notation numbers that get converted to numbers which is undesirable, but of scope of this change.

### DIFF
--- a/pre_commit_hooks/pretty_format_json.py
+++ b/pre_commit_hooks/pretty_format_json.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import re
 import sys
 from collections.abc import Mapping
 from collections.abc import Sequence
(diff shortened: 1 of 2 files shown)


asottile:
please discuss features before wasting time on them

I cannot accept a regex hack unfortunately and I don't think this is a good idea anyway -- invites more bikeshedding

adithyabsk:
> please discuss features before wasting time on them

Upstream was a curtesy, plan to use the fork to unblock myself. Understand if you cannot accept the change.

If you have suggestions on a better approach, I'm open to changing the implementation.

Spent some time to see if there were round trip stable `json` parsers but came up short. [`simplejson`](https://github.com/simplejson/simplejson) may be an option, but I didn't try it.

PR #1159, opened 2025-04-25
Find private keys within gzip-compresssed files

jwreschnig-utiq:
Unfortunately we have run into a situation where this was necessary but the precommit did not detect it.


asottile:
I don't think this is really in the spirit of the tool

you really shouldn't be checking in gzipped files to begin with

PR #1160, opened 2025-04-25
Adds note to Readme about `check-docstring-first`

Burhan-Q:
I bumped into the same issue raised in #159. I found a quick workaround after looking at the code, and updated the Readme with a note for anyone else looking for the same. 

Using VS Code (ver 1.99.3) I was able to add an attribute docstring wrapped in `()` and still get the docstring hint on hover, without `pre-commit-hooks` flagging the file.

image


asottile:
hmmm I don't really want to document that as I don't think your workaround is passing intentionally.  it's a quirk of how the hook uses the tokenizer

Issue #1078, opened 2024-07-31
end-of-file-fixer doesn't work correctly with CRLF endings


edobez:
`end-of-file-fixer` adds a LF before the last empty line, regardless of what are the EOL of the file.
On Windows, where typical EOL are CRFL, leaves the files with mixed endings.

edobez:
I added the following test line in end_of_file_fixer_test.py to reproduce the problem:
`(b'foo\r\nbar', 1, b'foo\r\nbar\r\n')`.

asottile:
please search the issue tracker

edobez:
I guess you mean this https://github.com/pre-commit/pre-commit-hooks/issues/326 issue.
The problem solved by that issue and related PR is to correctly detect CRLF endings, but it doesn't solve the problem stated here: in a  CRLF-ended file, we want that even the **last line** uses CRLF.

PR #1162, opened 2025-05-07
Handle CRLF and mixed-ending files

edobez:
Changes for `end-of-file-fixer`.

Refactor the handling of line endings to support files with CRLF and mixed endings.
In the first case, it will add a CRLF for the last line.
In the second case, it will default to LF for the last line.


edobez:
Refers to discussion in #1078.

asottile:
reading the whole file into memory defeats the purpose of this hook

asottile:
both

edobez:
Ok, so what about the solution of not caring about mixed line endings?

edobez:
I pushed that solution into my branch.

PR #1163, opened 2025-05-08
Handle CRLF and CR EOL in end-of-file-fixer

edobez:
Second attempt :-)

Correctly adding last-line ending based on the one detected at the first line.
Basically fixes the behavior for CRLF and CR based files.

### DIFF
--- a/pre_commit_hooks/end_of_file_fixer.py
+++ b/pre_commit_hooks/end_of_file_fixer.py
@@ -20,9 +20,13 @@ def fix_file(file_obj: IO[bytes]) -> int:
     last_character = file_obj.read(1)
     # last_character will be '' for an empty file
     if last_character not in {LF, CR} and last_character != b'':
+        # Check if file uses CRLF endings
+        file_obj.seek(0, os.SEEK_SET)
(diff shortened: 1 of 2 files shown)


pre_commit_hooks/end_of_file_fixer.py
```diff
@@ -15,13 +19,22 @@ def fix_file(file_obj: IO[bytes]) -> int:
         return 0
     last_character = file_obj.read(1)
     # last_character will be '' for an empty file
-    if last_character not in {b'\n', b'\r'} and last_character != b'':
+    if last_character not in {LF, CR} and last_character != b'':
+        # Look at first line to determine line ending
+        file_obj.seek(0, os.SEEK_SET)
+        first_line = file_obj.readline()
```
asottile, line 25: this will read the whole file into memory for `\r`-delimited files -- and will read the whole file into memory if it has no newlines

asottile (commented):
like I said -- I don't think this is fixable without significantly regressing performance so I'd rather just say it's unsupported

perhaps instead we should back out *any* handling of `\r`

PR #1173, opened 2025-06-21
implement fluent (ftl) check

sylvestre:
Fluent is file format used by Firefox and other programs for translation strings.


asottile:
please discuss features before spending time on then especially new hooks

generally these need to be universally useful to be added here and this is very specific

you're probably best to make your own repo

PR #1186, opened 2025-08-05
file_contents_sorter: add --group-cases-together -- a better case-insensitive sort

programmerjake:
The new `--group-cases-together` option groups lines that differ only in case together, so e.g.:
```
c
b
a
B
```
is sorted to:
```
a
B
b
c
```
instead of:
```
B
a
b
c
```
It also works with `--unique`.

The `--ignore-case` option keeps lines that differ only in case in the original sorted order, whereas the new option will sort them (and remove duplicates if you pass `--unique`).

I added a new option instead of changing `--ignore-case` for backward compatibility with existing commit hooks so it doesn't break CI.

better naming suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,8 @@ Note that this hook WILL remove blank lines and does NOT respect any comments.
 All newlines will be converted to line feeds (`\n`).
 
 The following arguments are available:
-- `--ignore-case` - fold lower case to upper case characters.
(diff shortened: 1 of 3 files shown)


asottile:
please discuss features first before spending time on them

I'm not really looking to add more options especially when I can't possibly think of a name for them or explain in simple terms how they work so I'm going to decline this sorry

PR #1204, opened 2025-10-17
Add --exclude to large files filter

tmi:
We often don't want to add large files by mistake, but there are some cases where this is legit, like large uv locks or jupyter notebooks. But they are not that large to warrant git-lfs, and we don't want to increase the 500 kB limit globally because of those

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Add this to your `.pre-commit-config.yaml`
 #### `check-added-large-files`
 Prevent giant files from being committed.
   - Specify what is "too large" with `args: ['--maxkb=123']` (default=500kB).
+  - Optionally exclude glob-like patterns with `args: ['--exclude=uv.lock,examples/*ipynb']`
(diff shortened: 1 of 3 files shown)


tmi:
to whom it may concern -- when testing this, I've run into a curious case, maybe a bug. I see in most places just `main()` is invoked, like https://github.com/pre-commit/pre-commit-hooks/blob/89b1f940f05226f188ae4b5dcf09d041612854de/pre_commit_hooks/check_added_large_files.py#L92, which however does not convert `sys.argv` into the first main's argument, and those then get lost. This happens even with explicit invocations of `pre-commit run`. Is that a known bug, or do I misunderstand something?

asottile:
there's no need. this is built in to pre-commit 

search the docs or the issue tracker next time please!

tmi:
ah I see, there is a general purpose `exclude:`, my bad, apologies. But @asottile  do you have any thoughts regarding the `argv` not being passed? Thx

PR #1212, opened 2025-11-19
Check for 'print' debug statement

AadityaSubedi:
Checks `print` statement calls as a debug statement.


asottile:
no thanks.  print isn't necessarily a debug statement

PR #1248, opened 2026-04-09
Fix requirements-txt-fixer ordering for --index-url and --extra-index-url

### LINKED ISSUE #612: `requirements-txt-fixer` ordering breaks using --index-url and --extra-index-url together in requirements.txt

Eric19881020:
## Summary
This PR fixes the ordering behavior in `requirements-txt-fixer` when both `--index-url` and `--extra-index-url` are present in `requirements.txt`.

## What changed
- added a regression test covering the `--index-url` / `--extra-index-url` ordering case
- updated the sorting logic so `--index-url` is ordered before `--extra-index-url`

## Why
Fixes #612.

## Notes
I’m a new contributor, so please let me know if you’d prefer a different test structure or implementation approach.

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -36,6 +36,15 @@ class Requirement:
 
         return name[:m.start()]
 
+    @property
+    def sort_key(self) -> tuple[int, bytes]:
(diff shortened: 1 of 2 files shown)


asottile:
sorry we don't accept ai slop

PR #1249, opened 2026-04-09
Fix requirements-txt-fixer reordering pip options

### LINKED ISSUE #612: `requirements-txt-fixer` ordering breaks using --index-url and --extra-index-url together in requirements.txt

Bahtya:
## Summary

Fixes #612

The `requirements-txt-fixer` hook was alphabetically sorting all lines in `requirements.txt`, including pip options like `--index-url` and `--extra-index-url`. This caused `--extra-index-url` to be sorted above `--index-url` (since "extra" < "index" alphabetically), which breaks pip's index resolution order.

## Changes

- **`pre_commit_hooks/requirements_txt_fixer.py`**: Added a check in `Requirement.__lt__` so that when both requirements being compared start with `--` (i.e., are pip options), they maintain their original relative order instead of being sorted alphabetically. Regular package requirements continue to be sorted alphabetically.

- **`tests/requirements_txt_fixer_test.py`**: Added 3 test cases:
  - `--index-url` before `--extra-index-url` stays in order (PASS)
  - `--extra-index-url` before `--index-url` stays in order (PASS, original order preserved)
  - Packages are still sorted while pip options remain in place (FAIL triggers sort, pip options unchanged)

## Test plan

- [x] All 416 existing tests pass
- [x] 3 new test cases cover the reported bug
- [x] Verified that `--index-url` / `--extra-index-url` ordering is preserved
- [x] Verified that regular package sorting still works correctly alongside pip options

Bahtya

### DIFF
--- a/pre_commit_hooks/requirements_txt_fixer.py
+++ b/pre_commit_hooks/requirements_txt_fixer.py
@@ -45,6 +45,14 @@ class Requirement:
         elif requirement.value == b'\n':
             return False
         else:
+            # if both are pip options (start with --), maintain original
+            # order to avoid breaking semantic ordering
+            # (e.g. --index-url must come before --extra-index-url)
+            if (
+                self.name.startswith(b'--') and
+                requirement.name.startswith(b'--')
+            ):
+                return False
             # if 2 requirements have the same name, the one with comments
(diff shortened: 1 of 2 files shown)


asottile:
ai slop prs are not acceptable

PR #1252, opened 2026-05-04
Correct argv parameter type

marinelay:
I recently noticed that **typeshed** has widened the type of `args` parameter in `argparse.parse_args` method from **Sequence[str]** to **Iterable[str]** ([typeshed#15613](https://github.com/python/typeshed/pull/15613)).

I believe it’s necessary to update the types of variables passed to parse_args to stay consistent with this change.


asottile:
you believe incorrectly unfortunately.  Sequence is compatible with Iterable

PR #1257, opened 2026-06-01
Allow sorting file contents on windows machines

TomsCodingCode:
I ran into this issue, because in one of our CI/CD pipeline's we run the pre-commit checks on a windows image, which would result in a puzzling output from the `file_contents_sorter` hook.

Before the hook would perform a `.rstrip(b'\n\r')` on all lines, removing the CR character, and then join the lines together, only using `'\n'`, resulting in white space differences that are hard to spot.

With this PR the newline sequence of the first line is used to determine how to strip and join the lines before and after sorting.

Added unit tests to ensure consistent behaviour on windows.

### DIFF
--- a/pre_commit_hooks/file_contents_sorter.py
+++ b/pre_commit_hooks/file_contents_sorter.py
@@ -29,18 +29,21 @@ def sort_file_contents(
     unique: bool = False,
 ) -> int:
     before = list(f)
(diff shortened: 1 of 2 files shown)


asottile:
this feature has been declined already.  you can read the discussion and recommendations if you search the issue tracker for it

PR #1263, opened 2026-06-24
Add remove-em-dash hook

JakubJablonski2-TomTom:
New fixer hook that replaces UTF-8 em-dashes (U+2014) with a plain hyphen (-), modeled on the trailing-whitespace hook.

- pre_commit_hooks/remove_em_dash.py: the fixer (binary-safe, UTF-8 only)
- tests/remove_em_dash_test.py: full coverage of fix and no-op cases
- registered in setup.cfg, .pre-commit-hooks.yaml, and README.md


asottile:
- we don't accept features without prior discussion
- we don't accept ai slop
- this doesn't seem very useful anyway

PR #1264, opened 2026-06-28
Add py.typed

zedzhen:
Add py.typed ([PEP 561](https://peps.python.org/pep-0561/), [typing documentation](https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information))


asottile:
there is no intended python api.  using any functions is at your own risk and will break without notice

PR #1275, opened 2026-09-21
check-yaml: report invalid UTF-8 instead of raising a traceback

0xacee:
## Problem

`check-yaml` opens each file as UTF-8 text, but the parse is only guarded against `ruamel.yaml.YAMLError`. A file that is not valid UTF-8 (for example latin-1) makes `load_fn` raise `UnicodeDecodeError`, which escapes `main()`:

```console
$ printf 'key: caf\xe9\n' > latin1.yaml
$ check-yaml latin1.yaml
  ...
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 8: invalid continuation byte
```

The exception aborts the loop, so any remaining files are never checked, and the user gets a stack trace instead of a diagnostic.

The sibling hooks already handle this cleanly: `check-json` catches `ValueError` (a superclass of `UnicodeDecodeError`), and `check-toml` gets a `TOMLDecodeError` from `tomllib`.

## Change

Catch `UnicodeDecodeError` next to `YAMLError` and report `filename: invalid UTF-8: ...`, so the hook marks the file as failed with a readable message and still checks the rest of the arguments.

## Testing

- New regression test `test_main_non_utf8_file` fails on `main` (`UnicodeDecodeError` propagates) and passes with the fix.
- `pytest tests/` â€” 414 passed, 4 xfailed.


asottile:
we don't accept ai slop

